PR #28, opened 2016-11-08
Device configuration application

kb1lqd:
I created a device configuration application for Faraday that queries and updates a local Faraday devices internal Flash memory configuration. An INI file contains the configuration to be loaded into a Faraday device. Along with the application creation I updated the device configuration module in the FaradayIO module to follow the design intent and other updates deemed worthy.

Device Configuration Application:
- GET '/' : Get's the queried devices configuration information. Returns a JSON dictionary of the data.
- POST '/' : Load the .INI file configuration into the local Faraday device respectively

FaradayIO:
- Updated device configuration to include a GPIO port 5 default bitmask
- Changed struct packet creation format string to contain actual 'x' padding bytes instead of blank strings
- Added a full received memory packet parsing struct format string
- Added packet extraction from encapsulation routine

Proxy:
- Removed the 164 byte packet check that made no sense and was incorrectly created. This blocks all but a select few transmissions to Faraday if implemented due to the 128 byte fixed frame and the BASE64 encoding to 164 bytes being very dependent on the encoded data.

To-DO:
- Update GET to also load received device configurations into the .INI file

### DIFF
--- a/.idea/Faraday-Software.iml
+++ b/.idea/Faraday-Software.iml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
+  <component name="NewModuleRootManager">
+    <content url="file://$MODULE_DIR$" />
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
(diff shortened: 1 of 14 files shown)


kb1lqd:
Looks good, pulling in. Might need to update Sphinx documentation a bit in FaradayIO device Config or new application later.

Issue #47, opened 2017-01-04
Telemetry Parsing Packet Device Debug Returns String Not Integer


kb1lqd:
The parsing module returns a string for boot counter when it should be an integer. Due to structure parsing no conversions are needed as integers are returned by default.

```python
dictionaryData = {'BootCounter': str(parsed_packet[0]),
                          'ResetCounter': int(parsed_packet[1]),
                          'BrownoutCounter': parsed_packet[2],
                          'Reset_NMICounter': parsed_packet[3],
                          'PMM_LowCounter': parsed_packet[4],
                          'PMM_HighCounter': parsed_packet[5],
                          'PMM_OVP_LowCounter': parsed_packet[6],
                          'PMM_OVP_HighCounter': parsed_packet[7],
                          'WatchdogTimeoutCounter': parsed_packet[8],
                          'FlashKeyViolationCounter': parsed_packet[9],
                          'FLLUnlockCounter': parsed_packet[10],
                          'PeripheralConfigCounter': parsed_packet[11],
                          'AccessViolationCounter': parsed_packet[12],
```

Should be:

``` python
dictionaryData = {'BootCounter': parsed_packet[0],
                          'ResetCounter': parsed_packet[1],
                          'BrownoutCounter': parsed_packet[2],
                          'Reset_NMICounter': parsed_packet[3],
                          'PMM_LowCounter': parsed_packet[4],
                          'PMM_HighCounter': parsed_packet[5],
                          'PMM_OVP_LowCounter': parsed_packet[6],
                          'PMM_OVP_HighCounter': parsed_packet[7],
                          'WatchdogTimeoutCounter': parsed_packet[8],
                          'FlashKeyViolationCounter': parsed_packet[9],
                          'FLLUnlockCounter': parsed_packet[10],
                          'PeripheralConfigCounter': parsed_packet[11],
                          'AccessViolationCounter': parsed_packet[12],
```

kb1lqd:
Committed in branch: https://github.com/FaradayRF/Faraday-Software/commit/e578ba84dfdd16333338f99a3d4761426fef7ef0

kb1lqd:
Created full branch to just fix this bug: 1c27050c4a0bef182f4e8b3cc1d1b839a8009366

kb1lqd:
Fixed in pull commit dad179864cdc8e93b534a4b69039b4a8efd9dcab

PR #48, opened 2017-01-04
Fixing Telemetry Parsing Packet #2 Error

kb1lqd:
Fixing bug #47 that incorrectly returned a string "0" instead of integer 0 and caused device testing check errors.


kb1lqc (approved):
Simple bug. Fix looks good as dictionary items are alrady ints.

PR #54, opened 2017-02-05
Device testing

kb1lqd:
- Added support specifically to device testing needs
- - Device functional test scripts to checkout hardware
- Bug fixes
- - Update ADCx parsing to correct "VCC" definition
- -


kb1lqc (approved):
Readme.md is bad an not relevant but we will get rid of that on a future commit.  Some INI file changes are unnecessary and we should figure out a working path forward. However, seems OK.

PR #55, opened 2017-02-05
Unboxing

kb1lqc:
Moved all applications into Applications folder and also cleaned up main readme.md as well as some start to telemetry and APRS documentation.


kb1lqd (changes requested):
Please remove the inclusion of your APRS login passcode in /APRS/login.ini.

Can we use a generic non-real callsign for all .ini config files such as /Telemetry/telemetry.ini so that we can default to a known OK one beside swapping between our two call's?

In tutorials can we use default generic callsign instead of our own? I've used NOCALL (letter o not number 0) in the past and comment to update to your callsign in use. This regards /start/hello-world.py

kb1lqc:
@kb1lqd I performed your requested action items. Please review again and approve if you agree.

kb1lqd:
Actions performed, hellow-world didn't have generic call updated but punting as low-priority to merger.

kb1lqd (approved):
Approved.

PR #68, opened 2017-02-12
Patched Bug #63

### LINKED ISSUE #63: Unit reports invalid preset GPS altitude
When a unit without a GPS is configured with set coordinates, the GPS Altitude that is returned is invalid unless it has 4 decimal points.

Configuration:
DEFAULT_ALTITUDE = 185
Returns:
"GPSALTITUDE": "185\u0000\u0000\u0000\u0000\u0000", 

Configuration:
DEFAULT_ALTITUDE = 185.000
Returns:
"GPSALTITUDE": "185.000\u0000",

Configuration:
DEFAULT_ALTITUDE = 185.0000
Returns:
"GPSALTITUDE": 185.0,

kb1lqd:
Fixed #63 and updated other similar items.

* Added length test status and check for global failure to deviceconfig.py so that data too long errors (could update to truncation later if wanted)
* Added other generic length checks for good practice
* Added custom padding packet creation function in commandmodule.py to pad data with specific bytes
* Force all Lat, Lon, Alt to floats and then convert back to string (all NMEA strings contain decimals)
* Padding function on Lat, Lon, Alt add 0x30 (ASCII 0) to each unused byte after data. This matches intended GPS NMEA DMM format in ASCII.


Applications/Telemetry/telemetry.ini
```diff
@@ -9,4 +9,4 @@ schemaname=db.sql
 [telemetry]
 units=1
 unit0call=REPLACEME
-unit0id=REPLACEME
```
kb1lqc, line 12: I removed your callsign

Applications/deviceconfiguration/faraday_config.ini
```diff
@@ -12,14 +12,14 @@ BOOT_FREQUENCY_MHZ = 914.5
 BOOT_RF_POWER = 20
 
 [GPS]
-DEFAULT_LATITUDE = 012345678
```
kb1lqc, line 15: You completely deleted DEFAULT_LATITUDE... make sure to test code prior to pull request...

Applications/deviceconfiguration/faraday_config.ini
```diff
@@ -12,14 +12,14 @@ BOOT_FREQUENCY_MHZ = 914.5
...
-DEFAULT_LATITUDE_DIRECTION = N
-DEFAULT_LONGITUDE = 0123456789
-DEFAULT_LONGITUDE_DIRECTION = W
-DEFAULT_ALTITUDE = 0
-DEFAULT_ALTITUDE_UNITS = M
+;NMEA DMM Format
+DEFAULT_LATITUDE_DIRECTION = N ; MAX LENGTH = 1 Byte
```
kb1lqc: While inline comments may have worked for you @kb1lqd they are not the standard from Windows which defined the INI file structure and therefore are not guaranteed to work.

https://stackoverflow.com/questions/1378219/do-standard-windows-ini-files-allow-comments#19550081

kb1lqc (dismissed):
I already changed a bunch of code thinking it would be simple fixes. However, When programming GPS lat/long I noticed that leading zeros were removed. You probably need to not convert to floats or at least directly (test then use string). 

This is an example APRS string which uses NMEA GPS data. I has leading zeros. You need to support this or we need to justify why not and pad zeros on the front in APRS only.

APRS,TCPXX*,qAX,CWOP-7:@130726z3729.30N/07959.48W_322/006g016t041r000p000P000b10160h40eMB32

kb1lqd:
@kb1lqc Yes you are right that the parsing on the RECEIVE side is dropping the leading zero(s) and this is unwanted. This was unintentional.

image

Printing the latitude string prior and post the conversion shows the obvious results.
image


I removed the float() and left the pre-packing input to a string of ASCII bytes and it worked except for altitude:

image


image

@kb1lqc Looks like the actual problem remaining in Telemetry.py for the Altitude being not text but a "Real" (integer?).

image

kb1lqd:
OK @kb1lqc I updated the functions to PREPEND instead of APPEND the ASCII zeros:

image


This makes sense from an NMEA formating perspective and also allows the use of INT instead of Float for Alt (i.e. 185 instead of forcing 185.0). In the end I think we'll be OK with this.

NOTE: Should we update the Telemetry.py SQL field for ALT from REAL to TEXT to display all characters (dropped zeros) as they are in the data going to the SQL databate?

kb1lqd:
I changed (did not commit) the telemetry SQL format for ALTITUDE to TEXT instead of real and it changed as expected:

image

@kb1lqc Do you want to save ALT as a text string like so or INT as before?

kb1lqd:
P.S. A nifty GPS converter tool for NMEA DMM to other formats located here: http://www.csgnetwork.com/gpscoordconv.html

kb1lqd:
NMEA Format: http://navspark.mybigcommerce.com/content/NMEA_Format_v0.1.pdf

image

Updates:
* I format GPS flash configuration data to be in formats
  * Lat: ddmm.mmmm (leading and trailing 0x30's ('0') padding bytes added automatically)
  * Lon: dddmm.mmmm (leading and trailing 0x30's ('0') padding bytes added automatically)
  * Alt: 8 Bytes forced to a Float (leading 0x30's ('0') padding bytes added automatically)

Note that _I deviate from the linked Specifical by also adding trailing pad bytes to Lat and Lon but justify this with clean userspace if only generic location is wanted._

kb1lqd:
Checking that input performs correct formatting:

## Correct Size Lat/Lon/Alt

image


image


## Short Lat/Lon/Alt:

Test cases worked.

image

image

## Long Lat/Lon/Alt

Correctly errors.

## Incorrect Format (no decimal)

image

Correctly failed

image

# Altitude Check

* Checks correctly for maximum of< 18km
* I chose to NOT support negative altitude at this time. I couldn't think of a nice way to format this in NMEA correctly and decided to move on. Sorry Death Valley, CA residents!
* Correctly checks length in bytes, converters to float as needed

kb1lqd:
@kb1lqc Please review again for implemented updates.

kb1lqc:
@kb1lqd did you commit? My Git log shows no new changes to New_Bug_63 have been made... Oh wait you branched again to Developer_Tutorial_Clean. Try to keep the branches the same. I'll have to merge Developer_Tutorial_Clean with this branch to keep the pull request valid... though it may pull in your tutorial updates..

kb1lqd:
Moved this merge over to #76

PR #70, opened 2017-02-13
New Faraday images and start tutorial formatting

kb1lqc:
Not requiring a reviewer for this as it's only a configuration file change and images being uploaded. Prior to @kb1lqd and I requiring more strict review cycles once project is a bit further along we will allow this in certain instances.


kb1lqc (commented):
Also removed a stray configuration.md file that must have somehow found its way into this folder...

PR #77, opened 2017-02-17
APRS Application Update & NMEA Format

kb1lqc:
The following updates were made:

- Correctly calculate position data for APRS Protocol Version 1.0 position packets. It appears to work well and correctly sizes position with leading and trailing zeros.
- Updated `aprs.ini` and `login-template.ini` with uppercase text
- Changed APRS-IS server to rotate.aprs.net so it rotates between servers. Seemed correct thing to do.
- Updates Applications/APRS/readme.md with basic configuration and use of APRS application.


kb1lqd:
Branch worked as described using non-GPS unit:

image

Tutorial was easy to follow.

Confirmed that manually changing location can cause APRS.FI to "speed" limit you, napkin calculations estimate around 299 792 458 m/s and above that invalidate you're packet. This is not a Faraday problem :)

image

kb1lqd:
@kb1lqc I confirmed it working with a GPS unit with GPS locked. I did not walk around but the computer connection was made and APRS data uploaded.

image

Data seems valid on APRS.fi and updating.

kb1lqd (approved):
Confirmed working program with both GPS and non-GPS unit (fixed location). Data verified on APRS.fi. Tutorial was good. Looking over code there were no obvious issues. INI files updated to uppercase as intended in future pushes.

Approved.

PR #83, opened 2017-02-18
Simpleconfig.py Updated to Show Programmed Data

kb1lqc:
I updated simpleconfig.py to query the Faraday radio over UART after programming to return its flash memory contents. The data is then parsed and printed very basically back to the user. It's not perfect and some bitmasks are not done too well but it's a lot better for now. I think it serves 90% of our need at the moment.

I did not worry about code formatting. This is a quick hack to get functionality ready for initial Faraday users this weekend.


Applications/deviceconfiguration/simpleconfig.py
```diff
@@ -40,14 +40,75 @@
...
+    unit_configuration_dict = cPickle.loads(b64_unit_json)
+    
+    #Print configuration values
+    print "\n************************************"
+    
+    print "POST CONFIGURATION UPDATE"
+    print "\nBASIC"
+    print "Unit Callsign:", str(unit_configuration_dict['local_callsign'])[0:unit_configuration_dict['local_callsign_length']]
+    print "ID:", str(unit_configuration_dict['local_callsign_id'])
+    print "CONFIGBOOTBITMASK:", bin(unit_configuration_dict['configuration_bitmask'])
```
kb1lqd: Since this is a binary print of the bitmask you should show all leading zeros: http://stackoverflow.com/questions/16926130/python-convert-to-binary-and-keep-leading-zeros

Applications/deviceconfiguration/simpleconfig.py
```diff
@@ -40,14 +40,75 @@
...
+    print "GPIO_P3:", bin(unit_configuration_dict['default_gpio_port_3_bitmask'])
+    print "GPIO_P4:", bin(unit_configuration_dict['default_gpio_port_4_bitmask'])
+    print "GPIO_P5:", bin(unit_configuration_dict['default_gpio_port_5_bitmask'])
```
kb1lqd: For all GPIO bitmasks:

Since this is a binary print of the bitmask you should show all leading zeros: http://stackoverflow.com/questions/16926130/python-convert-to-binary-and-keep-leading-zeros

Applications/deviceconfiguration/simpleconfig.py
```diff
@@ -40,14 +40,75 @@
...
+    print "\nGPS"
+    print "DEFAULT_LATITUDE:", str(unit_configuration_dict['default_gps_latitude'])
+    print "DEFAULT_LATITUDE_DIR:", str(unit_configuration_dict['default_gps_latitude_dir'])
+    print "DEFAULT_LONGITUDE:", str(unit_configuration_dict['default_longitude'])
+    print "DEFAULT_LONGITUDE_DIR:", str(unit_configuration_dict['default_longitude_dir'])
+    print "DEFAULT_ALTITUDE:", str(unit_configuration_dict['default_altitude'])
+    print "DEFAULT_ALTITUDE_UNITS:", str(unit_configuration_dict['default_altitude_units'])
+    print "GPS_BOOT_BIT & GPS_PRESENT_BIT:", bin(unit_configuration_dict['gps_boot_bitmask'])
+    print "\nTELEMETRY"
+    print "UART_TELEMETRY_BOOT_BIT & RF_TELEMETRY_BOOT_BIT:", bin(unit_configuration_dict['telemetry_boot_bitmask'])
```
kb1lqd: Since this is a binary print of the bitmask you should show all leading zeros: http://stackoverflow.com/questions/16926130/python-convert-to-binary-and-keep-leading-zeros

kb1lqd (changes requested):
Please update as noted to display bitmasks correctly.

kb1lqd:
Other than changes noted it works OK: 
image

Applications/deviceconfiguration/simpleconfig.py
kb1lqc: Updated

Applications/deviceconfiguration/simpleconfig.py
kb1lqc: Updated

Applications/deviceconfiguration/simpleconfig.py
kb1lqc: Updated

kb1lqc:
Updated bitmasks using string formatting. Output:

image

kb1lqc:
Seems like a quick check to get this pull request completed after my updates @kb1lqd! Thanks for the review.

kb1lqd (approved):

Issue #91, opened 2017-02-19
Telemetry .upper() in Stations Causing APRS to not start


kb1lqd:
I noticed that APRS will not start with the latest telemetry.py update. The APRS window crashes and programs fails to start.

![image](https://cloud.githubusercontent.com/assets/339144/23107528/77a52ad0-f6b3-11e6-84c3-196350b4bdf0.png)

Bug tracked to the "getStations()" telemetry call that APRS tries to use.

![image](https://cloud.githubusercontent.com/assets/339144/23107526/6fd5eb1e-f6b3-11e6-8c5e-102b4a89eff0.png)

Telemetry.py prompt reports this failure: 
![image](https://cloud.githubusercontent.com/assets/339144/23107547/d26a641c-f6b3-11e6-9442-4ca40dee491e.png)

I printed the callsign variable at line 357 and it was `NoneType`

I believe this is failing because the `.upper()` assumes a string but got `None`. This was recently merged and this situation (using APRS) probably wasn't vetted properly.

**Actions**

I'm going to push a simple pull request to remove the .upper() to allow the program to work. This may cause failure if the user doesn't enter their callsign in all UPPERCASE in the configuration(s).

kb1lqd:
Removing .upper() allows startup but no APRS upload yet. Waiting for GPS lock on this unit:

![image](https://cloud.githubusercontent.com/assets/339144/23107598/c09b778e-f6b4-11e6-822c-d35b798419eb.png)

I'm going to try to verify if I can solve this quickly without removing .upper() since it seems the getStations() isn't getting callsign first. The program is still failing non-gracefully.

kb1lqc:
Not sure if actually caused by upper() as you should a I'll be getting a callsign. Is something causing the callsign not to actually be in the dictionary?

kb1lqd:
@kb1lqc 

Well I think a different issue is that Telemetry/APRS fails hard if it doesn't see data. It should fail gracefully/not stop.

Tracking this down in aprs.py:

![image](https://cloud.githubusercontent.com/assets/339144/23107626/4bf3f07c-f6b5-11e6-9884-402f4c168dad.png)

![image](https://cloud.githubusercontent.com/assets/339144/23107634/60f9e4fe-f6b5-11e6-938e-f2847aebd48f.png)

Leads me to telemetry.py:

Looks like default if no URL argument supplied is `None`?

[...]

PR #93, opened 2017-02-21
Telemetry and APRS Callsign + Passcode Update

kb1lqc:
- Cleaned up telemetry.py upper() statements
- Forced upper() on telemetry callsign URL paramerter get() calls
- Changed callsign default value from "None" to "%" which passes an upper() call and is a valid SQL wildcard
- Removed SQL string generation for callsign = None
- No longer uses login.ini configuration file for APRS, APRS-IS callsign located in aprs.ini now
- Implemented a basic APRS-IS passcode generator. This automatically generates a passcode to protect against simple poor passcode entry and helps remove an extra configuration file.
- Checks for callsign= "REPLACEME" and warns user while not attempting a connection which will not work as intended.
- Changed telemetry rate to 60 seconds
- Changed socket.error exception checks over to ERRORIO in an attempt to gracefully fail better.
- Update APRS/readme.md to incorporate new configuration necessary to run with passcode generator.

This is related to #91 and should fix this issue. The code should be tested to work for all who properly configure the APRS application.


kb1lqc:
Please move this into smaller chunks. It's really really really hard to follow and understand with lots of little errors. If we want this to go out fast then make small bit-size pull requests that we can focus on and optimize to release one at a time.

Applications/APRS/aprs.ini
```diff
@@ -3,9 +3,10 @@ HOST=127.0.0.1
 PORT=8001
 
 [APRSIS]
+CALLSIGN=REPLACEME
```
kb1lqd, line 6: Why is this callsign here? Is this for passcode to APRS-IS?

kb1lqd: Oh is it this "No longer uses login.ini configuration file for APRS, APRS-IS callsign located in aprs.ini now"

Applications/APRS/aprs.py
```diff
@@ -260,7 +256,7 @@ def sendPositions(stations, socket):
                 try:
                     socket.sendall(positionString)
 
-                except socket.error as e:
+                except IOError as e:
```
kb1lqd, line 259: Why will this fail more gracefully than socket.error? What is IOERROR?

Applications/APRS/aprs.ini
kb1lqd: Disregard.

proxy/proxy.py
```diff
@@ -237,8 +237,8 @@ def proxy():
                 # Required
                 raise StandardError("Missing 'callsign' parameter")
             else:
-                # Ensure callsign value is a string and all uppercase
-                callsign = str(callsign).upper()
+                # Ensure callsign value is a stringe
```
kb1lqd: Non Blocking: Typo, should be string not stringe?

kb1lqd:
* Confirmed proxy forces callsign to UPPERCASE
* Confirmed telemetry forces callsign to UPPERCASE
* Confirmed APRS detects and soft errors on README left in INI file
* Confirmed APRS forces UPPERCASE
* Confirmed APRS works and connects, verified position reports updating on APRS.fi with a GPS enabled Faraday unit

Applications/APRS/readme.md
```diff
@@ -34,4 +32,4 @@ With these running simply run `aprs.py` and the application will startup. It con
  
 ## Correctly Running APRS Application
 ![APRS application running with fixed GPS location](images/APRS_Running_FixedPosition.exe.png "APRS Application")
-The application running above is nominal and connected to a `KB1LQC-1` Faraday radio through [Proxy](../../Proxy) which doesn't have a GPS. Since GPS Fix is therefore zero APRS simply warns that the unit has a bad GPS Fix. However, since we know it's configured with a fixed location in Flash memory this can be disregarded.
\ No newline at end of file
+The application running above is nominal and connected to a `KB1LQC-1` Faraday radio through [Proxy](../../Proxy) which doesn't have a GPS. Since GPS Fix is therefore zero APRS simply warns that the unit has a bad GPS Fix. However, since we know it's configured with a fixed location in Flash memory this can be disregarded.
```
kb1lqd: Blocking: Image not aligned as shown below. Text should be below the image not in-line.

image

kb1lqd (changes requested):
Most are questions/non-blocking. A few blocking.

Applications/APRS/readme.md
kb1lqc: Fixed

Applications/APRS/aprs.py
kb1lqc: `socket.error` is a [child of IOError](https://docs.python.org/2/library/exceptions.html#exceptions.IOError). This means IOError is a little more broad and sweeping. Additionally for some reason `socket.error` was not being found [when used in aprs.py](https://github.com/FaradayRF/Faraday-Software/issues/91#issuecomment-281220467). Using IOError fixed this as it's a built-in exception. I can investigate more as to why the socket module wasn't letting me do this but I figured this was an acceptable fix.

proxy/proxy.py
kb1lqc: Fixed

kb1lqd (approved):

Issue #81, opened 2017-02-18
Update limit checks for GPS altitude


kb1lqc:
Altitude is checked up to 17999.90 meters, however, we should be able to program static faradayconfig.ini files during Faraday configuration up to 17999.99.

Problem is in update_gps() in Faraday_Proxy_Tools/FaradayIO/deviceconfig.py and includes the check as well as the warning response text when the check fails.

kb1lqc:
Thanks @rtucker for the help!

PR #100, opened 2017-02-21
Fix altitude bounds check (#81)

rtucker:
Permit altitudes up to 17999.99.  Move the limits to be with
the rest of the hardcoded values.

Fixes issue #81


Faraday_Proxy_Tools/FaradayIO/deviceconfig.py
```diff
@@ -285,7 +285,7 @@ def update_gps(self, gps_boot_bitmask, latitude_str, latitude_dir_str, longitude
 
         else:
             print "ERROR: GPS string(s) too long OR NMEA DMM formatting incorrect"
-            print "ERROR: Altitude must be %g-%g" % (MIN_ALTITUDE_METERS, MAX_ALTITUDE_METERS)
+            print "ERROR: Altitude must be %g-%g" % (self.MIN_ALTITUDE_METERS, self.MAX_ALTITUDE_METERS)
```
kb1lqc: Hey @rtucker thanks for the pull request! How about we use Python .format() instead of % for string variable and formatting? This would correspond to the rest of Faraday Software + is the preferred method as of Python 2.6. It's also the preferred method for Python 3.0 if we move towards that down the line.

kb1lqc (changes requested):
Simple python preference and maintainability change preferred but otherwise this looks good. Make the update and I'll pull it in!

rtucker:
I agree, good call.  I'll fire up the laptop and fix this :-) -rt

rtucker:
Ah fudge, got distracted and f'd up my branch.  Grumble...

Will open a new clean pull request.

Issue #98, opened 2017-02-21
Add User Dependent .INI files to Gitignore 


kb1lqc:
All our configuration INI files are not included in the [.gitignore](https://github.com/FaradayRF/Faraday-Software/blob/master/.gitignore) file. This is because we want to track the files and they were under heavy development. However, we are now slowing down on these files changing and it is relevant to remove them so we do not accidentally commit them with our own serial ports and callsigns.

When these files are updated we can simply force Git to add them for that single special commit.
`git add --force my/ignore/file.foo`

## Tasks

- [x] Ensure all current .INI files in [Proxy](https://github.com/FaradayRF/Faraday-Software/blob/master/proxy), [Device Configuration](https://github.com/FaradayRF/Faraday-Software/blob/master/Applications/deviceconfiguration), [Telemetry](https://github.com/FaradayRF/Faraday-Software/blob/master/Applications/Telemetry), and [APRS](https://github.com/FaradayRF/Faraday-Software/blob/master/Applications/APRS) use "REPLACEME" for `CALLSIGN`, `NODEID`, and `COM`
- [x] Update [.gitignore](https://github.com/FaradayRF/Faraday-Software/blob/master/.gitignore) to ignore the specific INI files in these documents which change often and should be ignored since they change with every user (callsign, nodeid, etc).

kb1lqc:
Checking the `.gitignore` addition and the file seems correct in #110 . You correctly implemented the issue fix. However, it does not actually cause Git to ignore the files since we're already tracking them. [Source](https://help.github.com/articles/ignoring-files/).

Actually, look at: https://stackoverflow.com/questions/1329291/ignoring-an-already-checked-in-directorys-contents#1330097

`git update-index --assume-unchanged path/to/file.txt`

This seems to be more along our needs. However I'm not totally sure that it would not require every developer to run this after checking out the project. @hdkmike any suggestions?

kb1lqc:
I think I could have written a better Issue ticket in hindsight. I don't think we absolutely need the .gitignore additions. The result we want is to have git keep a specific version of the file in the repository but not allow more changes unless a change to that files is specifically called out as being worthy of updating.

Maybe this is how the original idea works with `git add --force` though I'm not convinced right now. Once we determine the path forward on this I think #110 can be pulled in.

jbemenheiser:
This isn't a new problem.  There are several recommended solutions here.  (https://gist.github.com/canton7/1423106) 
[...]

Issue #94, opened 2017-02-21
Remove proxy-template.ini from Proxy


kb1lqc:
We no longer need to use this file. Simply remove it!

kb1lqc:
Fixed in PR #124

PR #110, opened 2017-02-24
Delete proxy-template.ini, Ignore .ini files

jbemenheiser:
File cleanup that addresses issues #94  and #98


kb1lqc (commented):
- Verified that `proxy-template.ini` was removed per #94.
- Verified that .ini files are in `.gitignore` for current applications and proxy.

Added a comment to #98 regarding best use of .gitignore or another method to use with the .ini files.

kb1lqc:
@jbemenheiser are you in-fact updated to master in the branch this commit is from? It appears to think some of the files like faraday config ini are changing from the old version to what is now in Master.

kb1lqc:
@jbemenheiser it appears this PR branch is quite behind master which can simply be updated with an [upstream fetch](https://help.github.com/articles/syncing-a-fork/). I tried this on my own computer with your branch it did not complain so I'd imagine it will be quite quick. Enough changed that not doing this muddies up what is actually being changed by your PR.

kb1lqc:
Closing as PR #124 addresses this PR with cleaner base files for easier merging.

Issue #112, opened 2017-02-25
Standardize INI Configuration File format


kb1lqc:
## Summary
We need to settle on a configuration format for this project with our use of `.ini` files. The INI configuration file format is and informal one so we can make up some of the rules as long as they play nice with the parser reading these files. Updating existing applications such as `Proxy`, `Telemetry`, `APRS`, and `DeviceConfiguration` are the only necessary folders to work in at this time.

## Problem Explanation
The standard I'd like to settle on is:
```
;this is a comment but there are no inline comments
[SECTION]
NAME=value
```
- Uppercase `[SECTION]` and `NAME` fields. Values can be upper or lowercase
- No spaces before or after `=` separating the `NAME` from the `value`
- Comments must be on their own line and cannot be inline comments

This issue ticket does not refer to any logging .ini configuration files which should be dealt with on a separate ticket because they will require some additional programming to change to uppercase.

#### Software 
N/A

#### Hardware
N/A

# Supporting Information
Files that currently need to be updated

- [proxy.ini](https://github.com/FaradayRF/Faraday-Software/blob/master/proxy/proxy.ini)
- [Telemetry.ini](https://github.com/FaradayRF/Faraday-Software/blob/master/Applications/Telemetry/telemetry.ini)
- [aprs.ini](https://github.com/FaradayRF/Faraday-Software/blob/master/Applications/APRS/aprs.ini)
- [deviceconfiguration.ini](https://github.com/FaradayRF/Faraday-Software/blob/master/Applications/deviceconfiguration/deviceconfiguration.ini)
- [faraday_config.ini](https://github.com/FaradayRF/Faraday-Software/blob/master/Applications/deviceconfiguration/faraday_config.ini)

jbemenheiser:
I think pull request #117 wraps up this issue, so this ticket can be closed.

kb1lqc:
Closing, This should be added to contributions.md or some other method of ensuring future INI files follow the same style.

PR #115, opened 2017-02-27
Updated faraday_config.ini to have valid GPS coordinates by default 

jbemenheiser:
Updated the file per the standard on #112 and set the latitude, longitude and altitude to values of zero.  Programmed one unit and checked the config readback to confirm the gps values were programmed.


kb1lqc (approved):
Simple and effective! This will make programming Faraday radios slightly less painful 😆. Thanks!

Issue #113, opened 2017-02-26
APRS Doesn't Forward RF Stations Reliably to APRS-IS


kb1lqc:
## Summary
It appears that the APRS application doesn't actually send out any data to APRS-IS regarding an RF node with much reliability. It also appears that when the RF update rate is set to 3 seconds or so it is more reliable but when the RF interval is 30 seconds only a single packet of several will make it.

## Problem Explanation
With one radio operating on a 9V battery and another operating attached to the computer as a base station it appears that Telemetry reliably receives data. An RF node set to 30 second telemetry transmission rate while operating with a GPS lock and 9V battery about 15 feet away appears to have made it into the database just fine at ~30 second intervals:

![image](https://cloud.githubusercontent.com/assets/331540/23336394/b15d020a-fb83-11e6-8dc1-66c92d76e207.png)

The problem doesn't appear to be RF or Telemetry getting into the database. I suspect one of the following:

- Telemetry `/stations` app route in Flask is returning bad data
- APRS application is not using `/stations` data correctly to obtain latest telemetry packet from `/telemetry`
- APRS application is not using `/stations` data correctly to generate an internal list of stations

#### Software 
Master, Running on Windows 8. Unsure if it's present on other environments but it's likely.

#### Hardware
>Let us know about what hardware you are using if applicable. This includes:
>- Faraday radio serial numbers
SN 7 and SN 40, also confirmed with several other users so unlikely just this unit. 
>- Computer operating system
Likely all

# Supporting Information

APRS took a long time to actually send and APRS-IS update between successful updates. The unit KB1LQC-2 was reprogrammed with a 30 second RF interval after 19:14:09 but was on for at least 10 minutes prior to the successful update at 19:37:18. There were roughly 9 times KB1LQC-2 should have been successfully uploaded to APRS-IS

![image](https://cloud.githubusercontent.com/assets/331540/23336425/19229b42-fb85-11e6-8084-e6d52cedb0a3.png)

kb1lqc:
Interesting. Using a 30 second transmission and starting with a fresh telemetry database the first APRS transmission resulted in a `/stations` result of just the local node KB1LQC-1. The second APRS update 60 seconds later resulted in just the KB1LQC-2. Following that we see it's just KB1LQC-1 again...

```
C:\Python27\python.exe C:/Users/Bryce/Documents/GitHub/kb1lqc-software/Applications/APRS/aprs.py
[...]

PR #118, opened 2017-02-28
Issue113 Fixes

kb1lqc:
This PR addresses the symptoms in #113 
- Changed node symbole to an actual "node" with the "n" APRS symbol
- Added string to print how many stations are being logged for each update
- Removed printing of GPS Fix warning and any unnecessary data
- Added a logger.info() for each APRS telemetry packet function. This is for identification of future improvements. It's not necessary right now but will help make program more stable and recover from errors gracefully.
- Added programmingerror exception to SQL insertion function. This helps fail gracefully with bad data input
- Added rollback() to exception to allow database to work after exception.
- Fixed up some comment wording
- Fixed checking for callsign="%" in query DB fuction which caused a LIMIT=1 SQL query to be tacked onto all queries. This caused Issue #113.
  - If callsign is specified then the callsign is searched and returned stations are only nodes with that callsign. When no callsign is specified the QSL query is kept generic.

This code has a few known deficiencies
- Does not gracefully handle network errors and try to reconnect #119
- Telemetry will cause an error if no stations are returned. This is both a deficiency of APRS for not handling it well and telemetry for also not handling zero station /station API calls well #120.


kb1lqc:
OK @kb1lqd this is ready for review

Applications/APRS/aprs.ini
```diff
@@ -16,7 +16,7 @@ DATATYPEIDENT=!
 DESTADDRESS=GPSFDY
 SYMBOLTABLE=/
 ALTSYMBOLTABLE=\
-SYMBOL=N
+SYMBOL=n
```
kb1lqd, line 19: This is good, I was wondering why we were showing up as NTS stations...

Applications/Telemetry/telemetry.py
```diff
@@ -587,16 +595,20 @@ def queryStationsDb(parameters):
     sqlEnd = "GROUP BY SOURCECALLSIGN, SOURCEID ORDER BY EPOCH DESC"
 
     # detect if callsign/nodeid provided, return the last time it was heard
-    # Since a callsign was specified, simply search the entire db for it
-    # and return the last epoch time it was heard.
-    timeTuple = (0, time.time())
+    if parameters["CALLSIGN"] != "%":
```
kb1lqd, line 598: Seems to work fine. APRS is working with multiple stations. 1 Local and 1 Remote RF.

Applications/APRS/aprs.ini
kb1lqd: Verified that intended node type is shown on APRS.fi

kb1lqd (approved):
Reviewed code = Looks good
Ran APRS with units = All units local and remote verified on APRS.fi.

PR #140, opened 2017-03-31
Start fixing other issues reported by flake8

reillyeon:
Flake8 supports a large number of coding style checks. In the first commit here I explicitly disable all of the ones that cause it to flag issues across the project. Each subsequent commit then reenables one set of checks and fixes all the issues those checks flag.


kb1lqc:
I reviewed and this looks great. Thank you for the atomic commits! @kb1lqd this is a great example of clear and easy to review committing with each commit message giving a solid explanation of exactly what happened. All I want to do prior to pulling this in is run it through hardware. Otherwise I didn't find anything that makes me think this should not work.

kb1lqc:
Tested `Proxy`, `Telemetry`, `DeviceConfiguration`, `Simpleconfig`, and `APRS` with this PR and all seems well!

kb1lqc (approved):
Tested with hardware and all seems well.

PR #141, opened 2017-04-02
Add Flash Configuration Read Program To Device Configuration Application 

kb1lqd:
This pull request adds a modified "simpleconfig.py" that reads a locally connected units flash configuration and displays it without reprogramming the unit. This is useful for reading a the configuration of a Faraday unit without needing to program the unit.


kb1lqc (approved):
Readconfig works fine and I removed use of cPickle and added in JSON module functionality. The readconfig script is definitely like simpleconfig and very basic/non-optimized but OK for use as it's more of a quick and dirty tool than good program.

PR #142, opened 2017-04-03
Use pytest to manage running flake8

reillyeon:
By switching to using pytest and the pytest-flake8 module unit testing and code style checking can be run in the same command. This should make it easier for developers on Windows since presubmit.sh was a Bash shell script.


kb1lqc (approved):
@reillyeon  I read up on pytest and [tools:pytest] fixed a conflict of using the [pytest] setup.cfg section. This makes sense and what you're doing is starting up pytest with the flake8 configuration command line options.

Also, you removed `presubmit.sh` but I don't see a new `py.test` file. Should I assume the script yml runs py.test which is actually calling the python module pytest with the command line options instead of running a local custom script?

I'd like this questions answered but I'm also pretty sure I'm right so I am approving. I'll wait and let you merge if if you agree.

reillyeon:
`py.test` is the command added by installing the pytest library so no script is needed anymore. Unfortunately I don't have write access to the repository so even with approval I still need you to hit the merge button.

Issue #108, opened 2017-02-24
Add Functionality to Proxy For Logging Data to SQLite3 Database


kb1lqc:
## Summary
We should be able to "fake" being a Faraday radio by looking like a COM port sending valid UART data. This would allow us to automate testing and iterate without hardware. It would also allow people without Faraday hardware to use the software and potentially help contribute back to the project.

## Problem Explanation
The following Stack Overflow summarizes the options nicely:
https://stackoverflow.com/questions/1605721/faking-an-rs232-serial-port

The option also exists to not emulate at the COM port and actually tie in as a fake proxy. We could put a mode into Proxy that reads from a SQLite database or another predefined file to generate fake data.

#### Software 
This should be cross platform but we are largely developing on Windows at this time. I would love to see Windows, Linux, Max OS X support at some point.

#### Hardware
The goal is to remove the need for hardware!

# Supporting Information
N/A

kb1lqd:
@kb1lqc  Is this important? Over dev tutorials?

kb1lqc:
It is important, @jbemenheiser seemed interested in taking this on slowly over time when I talked with him this weekend. Might make sense to do less of a COM port simulator and more of an addition to Proxy which saves BASE64 strings of at least telemetry (for starters) so it can be served up in a "debug" mode of Proxy that doesn't connect to serial COM and just sends what's in the database.

jbemenheiser:
I am interested in working on this issue.  My understanding of the task was to create a mode within Proxy that can send out data for other applications without hardware connected.  

Hardware Simulation Mode:
* Make a database with sample data in BASE64 strings that has changing values
* Within proxy.py, add functions to read in data from the database and send it out at the configured rate
* Within proxy.sample.ini configure simulation mode and datarate

I would probably still start with running through the developer tutorials first.  I don't know the code well, so I need to get up to speed there.  Having this task in mind as I go through the tutorials will give me an external point of reference.

b00klegger:
+1 @jbemenheiser
I have (like others I'm sure) finally cobbled together something in proxy which would seem to make Proxy run in a 'normal' way to use as a base for further development. Nothing as sophisticated as what has been proposed here of course. Ping me if you want some help.

kb1lqc:
FYI @b00klegger and others you can use the up-vote emoji which is used as an actual up-vote count on GitHub!

[...]

PR #144, opened 2017-04-05
Adding Logging Functionality to Proxy

kb1lqc:
Per #108 I implemented a method of saving data to an SQLite3 database. Largely this will be for testing Telemetry and APRS functionality. I used some methodology discussed with @b00klegger but have minimized the code additions and overall footprint of this code to remain in scope with **only logging data**. We will use separate efforts to actually test the code. This code is needed now to help flesh out how we are automating testing.

Essentially this code is extremely straightforward. I'd suggest stepping through the commits from beginning to end to see the thought process.

When enabled in `proxy.ini`, logging is performed to `log.db` in the proxy folder.

# Database Format
The database is setup in a simplistic manner with a `KeyID` that is unique and auto-increments. Then the Faraday `Port` is saved along with actual `BASE64` data. Lastly, every row is given a computer generated `EPOCH` timestamp indicating when the data was inserted into the database.

# Configuration
A simple boolean value in proxy.ini was added to control logging.

# Logging in Action
As shown below, logging is easily seen with DB Browser for SQLite
proxylog


Proxy/proxy.py
```diff
@@ -71,10 +72,20 @@ def uart_worker(modem, getDicts, units):
...
+
+                            # Check for Proxy logging and save to SQL if true
+                            if log:
+                                item["port"] = port
+                                sqlInsert(item)
                         except:
                             getDicts[unit][port] = deque([], maxlen=100)
                             getDicts[unit][port].append(item)
 
+                            # Check for Proxy logging and save to SQL if true
```
reillyeon: If we log `item` in either case why not move this to after the try/except block.

Proxy/proxy.py
```diff
@@ -351,10 +362,106 @@ def callsign2COM():
...
+        except sqlite3.Error as e:
+            logger.error("Sqlite3.Error: " + str(e))
+            return False
+
+    return True
+
+
+def sqlInsert(data):
+    """
+    Takes in a data tuple and inserts int into the Proxy SQLite table
```
reillyeon: int -> it

Proxy/proxy.py
```diff
@@ -351,10 +362,106 @@ def callsign2COM():
...
+        try:
+            conn = sqlite3.connect(db)
+
+        except sqlite3.Error as e:
+            logger.error("Sqlite3.Error: " + str(e))
+            return False
+
+        # Connect to database, create SQL query, execute query, and close database
+        try:
+            conn = sqlite3.connect(db)
```
reillyeon: `conn` was already opened in the try/except block above.

Proxy/proxy.py
```diff
@@ -351,10 +362,106 @@ def callsign2COM():
...
+            # Use connection as context manager to rollback automatically if error
+            with conn:
+                conn.execute(sql,sqlparameters)
+
+        except sqlite3.Error as e:
+            logger.error("Sqlite3.Error: " + str(e))
+            conn.rollback()
```
reillyeon: Using `conn` as a context manager will automatically call `rollback()` so it is not necessary here.

reillyeon (dismissed):

Proxy/proxy.py
kb1lqc: Good point. Fixed

Proxy/proxy.py
kb1lqc: Good catch! [Per the docs](https://docs.python.org/2/library/sqlite3.html#using-the-connection-as-a-context-manager). Removed

Proxy/proxy.py
kb1lqc: Removed. Good catch!

kb1lqd (approved):
Looks good.

Issue #148, opened 2017-04-07
Add CORS Support to Telemetry Application


kb1lqc:
## Summary
[CORS support](https://enable-cors.org/) is necessary to share data across ports on localhost. [Flask-CORS](https://pypi.python.org/pypi/Flask-Cors) solves this problem easily. We should implement it.

## Problem Explanation
In order for simple local sharing of JSON data across localhost ports the telemetry server must enable CORS support for cross-site access. This is because we jump ports. 

#### Software 
Faraday-Software

#### Hardware
N/A

# Supporting Information
If CORS isn't supported by Telemetry then a local javascript function running on port 80 will error when retrieving telemetry data on port 8001:

![image](https://cloud.githubusercontent.com/assets/331540/24788862/b1801e50-1b24-11e7-8f4c-05bd62d8c2f6.png)

This is as easy as implementing the "Simple Usage" example from the [Flask-CORS documentation](https://pypi.python.org/pypi/Flask-Cors).

![image](https://cloud.githubusercontent.com/assets/331540/24788914/04bf9a3c-1b25-11e7-9253-d91f35cb8390.png)

jsr38:
This looks straightforward.  I'll take it on.  Just brought up Telemetry and Proxy with a fake nodeid and callsign.  I'm trying to check that the CORS headers are correctly generated but as nothing gets sent back to the client I can't see.  Is there a simple way to initialise Telemetry such that it can return something in order for me to verify the headers.  I'm sure I can come up with something - I'm asking in case there is a well-known best way in the absence of Faraday hardware...

kb1lqc:
@jsr38 we are not quite yet able to run Faraday software without hardware. Issue #108 pulled in logging functionality to save BASE64 data from a Faraday radio so we can play it back but we haven't implemented the ability to play it back yet. Shortly!

For the time being I'm happy to quickly test. This Issue to add CORS support is literally two lines of code so it should be very straightforward and my description gives you 99% of what you need to know to implement it. Just find the appropriate spots in Telemetry.py to put the import and the CORS wrapper with the flask "app" descriptor and you're set.

This will help get you into the [fork and PR model](https://gist.github.com/Chaser324/ce0505fbed06b947d962) if you're not used to it either.

kb1lqc:
Merged in #149. Thanks @jsr38!

Issue #145, opened 2017-04-06
Create Simple Telemetry/Command Interface For HAB


kb1lqd:
## Summary
The Rose-Hulman amateur radio club is preparing a high-altitude balloon with Faraday on-board and they would like a simple telemetry and command interface. The prior GUI used in https://faradayrf.com/simple-faraday-balloon-launch/ is not currently working with our new modular software. Due to time priority a simpler interface is acceptable.

## Problem Explanation
A simple user interface to:
* Receive and display telemetry live
* Basic telemetry ADC scaling functionality to scale analog sensors
  * Better left for Telemetry?
* Command Faraday for balloon mission

Since proxy and telemetry are Flask based a basic command prompt interface or simple webpage output is acceptable for telemetry viewing.

For commanding a simple command prompt interface is acceptable.

#### Software 
Faraday Master branch

#### Hardware

Standard Faraday hardware

# Supporting Information

kb1lqc:
Looking at the code we flew with Faraday back in August 2016... it was intimately integrated with Proxy. No bueno. I really don't want to temporarily port that in. We've moved on quite readily.

It will likely be easy to provide an automatically updating jQuery based webpage to see telemetry. No frills. As for commanding, leaving that for command prompt reduces the jQuery and other work to integrate buttons with RESTful commands and therefor would simplify this more.

kb1lqc:
I've pushed a few updates to my simpleui branch: https://github.com/kb1lqc/Faraday-Software/tree/simple_web_ui

I now have basic telemetry (subset) code updating to a webpage automatically using JQuery and javascript. This is a good step towards no frills telemetry. This requires the Telemetry CORS update on #148.

Below is a quick screencapture of the interface.

![simpleui](https://cloud.githubusercontent.com/assets/331540/24791070/9034ebdc-1b2d-11e7-9c20-08547871c025.gif)

kb1lqc:
Current plan:
- [x] Scale telemetry in javascript to volts at the ADC 
- [x] Allow selection of station to be displayed
- [x] Add commanding support for basic IO commands
- [x] Provide buttons for commands
- [x] Provide simple Bootstrap styling

kb1lqc:
Getting there. Boostrap styled UI with some javascript logic to display time and speed nicely as well as detect incorrect input data. Next up simple commanding!

[...]

PR #149, opened 2017-04-11
Fix for issue #148 - add CORS support to Telemetry.

jsr38:
Add CORS support to Telemetry.  Have updated requirements.txt to include the new dependency.  Tested in as far as I can start the server with the same result as before but cannot (easily) currently check for the presence of CORS HTTP headers.

### DIFF
--- a/Applications/Telemetry/telemetry.py
+++ b/Applications/Telemetry/telemetry.py
@@ -22,6 +22,7 @@ import json
 
 from flask import Flask
 from flask import request
+from flask_cors import CORS, cross_origin
 
 # Can we clean this up?
 sys.path.append(os.path.join(os.path.dirname(__file__), "../../Faraday_Proxy_Tools/"))  #Append path to common tutorial FaradayIO module
@@ -146,6 +147,8 @@ def telemetry_worker(config):
 
 # Initialize Flask microframework
 app = Flask(__name__)
+# Enable CORS support
+CORS(app)
 
 
 @app.route('/', methods=['GET'])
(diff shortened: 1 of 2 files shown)


jsr38:
I see the CI build failed, rightly due to:  

/home/travis/build/FaradayRF/Faraday-Software/Applications/Telemetry/telemetry.py:25:1: F401 'flask_cors.cross_origin' imported but unused

To avoid this in future, what do I need to do to enable strict checking or to run the CI check itself before I submit?

kb1lqc:
@jsr38 you should run it locally prior before submitting the PR but totally OK right now. Yes I forgot to mention this will not be used until I pull in #145. Therefore I'll manually check to see if this is correctly implemented and will pull it immediately after #145 and cite the necessary merge.

kb1lqc (approved):
These look great! Will pull in once #145 is implemented and ready to pull into master.

**This branch will fail Travic-CI until #145 is merged which actually uses CORS**

jsr38:
Of course I at least ran it ;-)  I'd like to know how I can run it with the strict checks (warnings -> failure promotion) equivalent to the way CI invokes it, locally.  Something to do with PEP8?  I can probably figure it out but I'm wondering if there's a pattern.  Something about  FLAKE8-check(ignoring E262 E265 E266 E402 E501).  I'm not familiar with the CI setup.  Is it possible to get a suitably modified "py.test" to run locally?

kb1lqc (approved):
Tested this updated code with cross_origin removed using hardware and my simple user interface program which requires CORS and it works fine.

Issue #151, opened 2017-04-11
Telemetry Incorrectly Parses IOSTATE


kb1lqc:
## Summary
IOSTATE, GPIOSTATE, and RFSTATE are all sent down to indicate IO state of various parts of Faraday. The Telemetry Parser has a bug in it where it does not correctly parse IOSTATE and simply leaves it out. This should be correctly parsed and available in the SQLite database.

## Problem Explanation
IOSTATE is correctly parsed from the packet [here](https://github.com/FaradayRF/Faraday-Software/blob/master/Faraday_Proxy_Tools/FaradayIO/telemetryparser.py#L172) after being correctly unpacked from the struct with the "3B" on [this line](https://github.com/FaradayRF/Faraday-Software/blob/master/Faraday_Proxy_Tools/FaradayIO/telemetryparser.py#L18). It appears the error is located in createTelemetryList() [here](https://github.com/FaradayRF/Faraday-Software/blob/master/Applications/Telemetry/telemetry.py#L486) where IOSTATE is never added to the telemetry list.

This will require at a minimum fixes to the following areas:
- [createTelemetryList()](https://github.com/FaradayRF/Faraday-Software/blob/master/Applications/Telemetry/telemetry.py#L486): parse IOSTATE
- [sqlInsert()](https://github.com/FaradayRF/Faraday-Software/blob/master/Applications/Telemetry/telemetry.py#L532): Provide correct SQL parameter and insertion code to place into SQLite3 db
- [db.sql](https://github.com/FaradayRF/Faraday-Software/blob/master/Applications/Telemetry/db.sql): To add in an IOSTATE Telemetry column

#### Software 
Python Master software branch

#### Hardware
Any Rev D1 Faraday hardware will allow this to be observed. Once Telemetry replay is allowed this also would have been caught since that method uses BASE64 data which contains IOSTATE even if telemetry had missed it.

# Supporting Information
Current Packet Format for telemetry packets:
![pasted image at 2017_04_10 10_26 pm](https://cloud.githubusercontent.com/assets/331540/24895036/104d8258-1e43-11e7-91cc-e3fa13ab2744.png)

Current IOSTATE, RFSTATE, and GPIOSTATE format:
![pasted image at 2017_04_10 10_22 pm](https://cloud.githubusercontent.com/assets/331540/24895054/2cb240f0-1e43-11e7-949e-d21698a57b08.png)

[...]

PR #154, opened 2017-04-14
Parsing IOSTATE With Telemetry Module

kb1lqc:
Implementing simple fixes suggested by #151, this pull request helps fix a bug in the SimpleUI #145 where IOSTATE was never actually parsed into the SQLite database. This addition allows the eventual use of the data which was not able to be retained prior versions of telemetry.

Tasks to pull in
- [x] createTelemetryList(): parse IOSTATE
- [x] sqlInsert(): Provide correct SQL parameter and insertion code to place into SQLite3 db
- [x] db.sql: To add in an IOSTATE Telemetry column


kb1lqc:
> sqlInsert(): Provide correct SQL parameter and insertion code to place into SQLite3 db

This is performed automatically because in sqlinsert() the function simply adds the correct number of parameter '?' per items in the telemetry list and inserts them in the same order as telemetrylist. Therefor as long as the item is in telemetrylist it will be inserted into the database in the order that is specified.

kb1lqc:
OK @kb1lqd this is ready for review. SUPER simple.

Issue #152, opened 2017-04-11
Enabled Proxy Playback Mode


kb1lqc:
## Summary
#108 provided the ability for saving data from hardware into Proxy by means of a SQLite3 database. Now we need to "play" the database file back as if it was real hardware. This will allow us to send a `log.db` file to someone and have them play it as if they had hardware connected. It will also allow automated testing of Telemetry/APRS functionality at the minimum.

## Problem Explanation
Issue #108 created a `log.db` file that we should now allow the configuration of the `Proxy.py` file and config INI to skip setting up a serial port connection and instead open the database and serve each row. At first we should simply serve one row every `rate` seconds which should be set in the INI file. Eventually we may desire to automatically know the rate of the SQLite file. Simplicity is key first.

#### Software 
Faraday Software Master branch

#### Hardware
None, please use provided log.db

# Supporting Information
`log.db` SQLite database provided in zip file.
[log.zip](https://github.com/FaradayRF/Faraday-Software/files/913988/log.zip)

jsr38:
Right @kb1lqc , I've had a quick look and I propose the following:

- changes to configuration that look like this:  
```
[TESTDATABASE]
FILENAME=log.test.db
SCHEMANAME=db.sql

[PROXY]
LOG=0
UNITS=1
TESTMODE=1
TESTRATE=1
```

- Where `TESTMODE=1` causes the UART init to be skipped and the DB to be read and inserted on a deque periodically.

- In order to stay consistent with the current Proxy implementation, the playback packets and the deque should be identified as coming from a Station, or callsign-nodeid pair.  I don't think this information is encoded in the packet - please tell me if I'm wrong and where to look if the packet contents can be used to determine this.  I would suspect not in a layered protocol so somehow this info must be provided, perhaps with a `TESTCALLSIGN` and `TESTNODEID` in the config.

- I propose the deque simply overflow if no client is present to read packets.

- Think that use of an asynchronous timer might be better than dealing with the thread directly, although since the uart threads are not going to be started in this situation, it might be simplest to follow the existing pattern and create a `read_testdb_worker` function and sleep.

Let me know if this is along the right lines.

kb1lqc:
- Configuration file looks 👍!
- You are correct, the callsign information that refers to which station Proxy is sending data from (which all other applications connection to) is not sent with the packet. Each deque is the value of a dictionary {key:value} where the key is the `callsign-nodeid` of the radio connected via USB.
[...]

PR #155, opened 2017-04-15
Issue #152 - First pass at enabling Proxy playback of logged data.  

jsr38:
Reads at rate TESTRATE packets per second, appending to the relevant deque from [TESTDATABASE] when TESTMODE=1, otherwise functions as before.


Proxy/proxy.py
```diff
@@ -111,6 +111,70 @@ def uart_worker(modem, getDicts, units, log):
...
+    """
+    logger.info('Starting testdb_read_worker thread')
+
+    # Obtain the test callsign and nodeid and create a
+    # deque
+
+    # Obtain configuration properties
+    try:
+        testCallsign = proxyConfig.get("PROXY", "TESTCALLSIGN")
+        testNodeId = proxyConfig.get("PROXY", "TESTNODEID")
```
kb1lqc: Could simplify to use `proxyConfig.getint()` so it will force an int

jsr38: Sure thing... 👍

Proxy/proxy.py
```diff
@@ -393,6 +457,59 @@ def initDB():
...
+
+    except ConfigParser.Error as e:
+        logger.error("ConfigParse.Error: " + str(e))
+        return None
+
+    if not os.path.isfile(testDbFilename):
+        logger.error('Test database: {} not found. '.format(testDbFilename))
+        return None
+
+    con = None
```
kb1lqc: Did you mean `conn`?

Proxy/proxy.py
```diff
@@ -111,6 +111,70 @@ def uart_worker(modem, getDicts, units, log):
...
+
+    conn = openTestDB()
+    if conn is None:
+        return
+
+    cursor = sqlBeginRead(conn)
+    if cursor is None:
+        return
+
+    row = cursor.fetchone()
```
kb1lqc, line 159: Wouldn't you want to `fetchall()` and then iterate through all returned?

Proxy/proxy.py
jsr38: See subsequent commit.

Proxy/proxy.sample.ini
```diff
@@ -6,9 +6,16 @@ PORT=8000
...
 
+[TESTDATABASE]
+FILENAME=log.test.db
+
 [PROXY]
 LOG=0
 UNITS=1
+TESTMODE=0
+TESTRATE=1
+TESTCALLSIGN=TESTCS
```
kb1lqc: Change sample configuration file callsign & nodeid to `REPLACEME` since it's standard to need to replace them if you are to use it.

Proxy/proxy.py
jsr38: Well, it avoids assuming it'll all fit in memory.  Happy to do either.

Proxy/proxy.sample.ini
jsr38: Sure thing... 👍

Proxy/proxy.py
```diff
@@ -448,6 +565,7 @@ def sqlInsert(data):
 
 def main():
     log = proxyConfig.getboolean('PROXY', 'LOG')
+    testmode = proxyConfig.getboolean('PROXY', 'TESTMODE')
```
kb1lqc: Just noting that we should probably fail gracefully on this with a try/except on config read but I won't hold up the PR for this as I am pretty sure `Proxy` hasn't received attention to this at all anyways.

jsr38: I saw it done in other places but followed suit on this occasion.  I'm gonna touch the files anyhow so I'll do it for these two config reads.

Proxy/proxy.sample.ini
```diff
@@ -6,9 +6,16 @@ PORT=8000
 FILENAME=log.db
 SCHEMANAME=db.sql
 
+[TESTDATABASE]
+FILENAME=log.test.db
+
 [PROXY]
 LOG=0
 UNITS=1
```
kb1lqc, line 14: Just realized if we can have more than one unit in the test mode then we need to be able to identify which unit is which callsign-nodeid. We did this by using separate sections as shown elsewhere in this INI file. However that could start to get messy... Would something like `TESTCALLSIGN0` and similar be acceptable here?

Proxy/proxy.py
```diff
@@ -111,6 +111,70 @@ def uart_worker(modem, getDicts, units, log):
...
+
+    row = cursor.fetchone()
+
+    # Loop through each row placing each row into deque
+    while(row is not None):
+
+        port = row[1]
+        item = row[2]
+        try:
+            getDicts[unit][port].append(item)
```
kb1lqc: This needs to be a dictionary with key `"data"` as shown [here](https://github.com/FaradayRF/Faraday-Software/blob/master/Proxy/proxy.py#L71). Causes an error when I run `Telemetry` program with a log file.

kb1lqc (changes requested):
I left some comments, noted which ones were more suggestions than necessary but you absolutely must change the `item` in the test worker to be a dictionary with a key of data for this to work. I tested it in my local branch and it works like a charm once implemented!

kb1lqc:
Here's a quick view of the telemetry application showing test proxy data with NO hardware connected once the item["data"] error was resolved (I locally changed it for my testing):

telemetry-testproxy

Proxy/proxy.sample.ini
jsr38: Can run with that.  Multiple test units needs quite a bit more thought.  I'm slightly confused about why the port is written to the DB and not the callsign or nodeid.  At some later point in time, after a configuration has been changed for example, the log no longer makes sense.  Also, I could imagine that the DB log could be interspersed with packets from multiple nodeids and the only way to differentiate is currently port number.  If the callsign and nodeid were logged a single DB could play back traffic for a number of units.  I'm not sure about the design and why it is currently so that may not be what's desired.

Proxy/proxy.py
jsr38: OK didn't spot that.  Will change.  Thanks.

kb1lqc:
Just checked my telemetry database and noticed an interesting (not wrong) observation. Even though my test proxy and telemetry application are setup as `KB1LQC-10` the BASE64 data as it should be is only showing `KB1LQC-1` for both source and destination stations. This is expected because my log file attached to #152 only recorded my local Faraday radio `KB1LQC-1` sending UART data so this would be correct. I simply realized that the proxy/telemetry callsigns are not updated to nodeid 10 which is correct that it should NOT be but I for some reason was surprised to see this ;)

image

jsr38:
Yes, I did wonder about parsing the content of the message.  In fact, when I tested against Telemetry I incorrectly assumed the failure was due to the content not being parsed properly by the Telemetry side.  I think it was late in the day by then as I didn't look any further at it :-( 

Now that I can see the interpreted packets, does it make sense to parse the packet to append to the 'correct' deque?  I don't know anything about the network and it sounds like although you were surprised, you actually want to be able to force the locally connected callsign-nodeid...

Proxy/proxy.py
kb1lqc: Ohh you fetchone() at the end. OK I see how this works. That makes total sense. My bad!

Proxy/proxy.sample.ini
kb1lqc: @jsr38 Yeah this is an area where it would help if @kb1lqd and I documented it a bit better. We do need to spend some time on this type of documentation. The port number is essentially an eight bit number (0-255) that indicates a potential application on Faraday itself written in firmware. For example `port 5` is the telemetry application on Faraday so any data received from port 5 will be related to telemetry. Any 'port 2' data is intended for the command/control application that controls GPIO and other settings. Think of it as the UART COM is an "IP Address" and the `Port` is the one of many ports available on an IP address (i.e. 80 for HTTP, 22 for SSH, etc)... Similar concept

kb1lqc:
Yeah each deque dictionary is organized with a callsign-nodeid key so that probably makes sense to abide by.

kb1lqc:
An observation on variable/function naming schemes. We technically have [guidelines for that now](https://github.com/FaradayRF/Faraday-Software/blob/master/CONTRIBUTING.md#python). @kb1lqd and I were not consistent in this when we started the project but we should enforce it more. Your new code doesn't follow this and I wouldn't hold the old code to that level of scrutiny on this PR (likely another PR) but we should be doing our best to abide by it.

I don't want to bog you down with this formatting stuff, I might update it for you!

Proxy/proxy.py
```diff
@@ -161,7 +162,8 @@ def testdb_read_worker():
     while(row is not None):
 
         port = row[1]
-        item = row[2]
+        item = {}
+        item["data"] = row[2]
         try:
             getDicts[unit][port].append(item)
```
kb1lqc: Yup and right here getDicts[unit][port] is correctly adding the data to a dictionary of dequeues for each station just like the normal Proxy. Looks good!

kb1lqc:
On second thought proxy needs to just be updated at once for the function names and stuff. Lets really only focus on new code with this unless you would really like to go back and fix this @jsr38. I'll approve but leave it up to you for this.

kb1lqc (approved):
Looks good. I've noticed a few items that should be noted:
- Does not have ability to "repeat" log file. Makes sense and wasn't actually a requirement but my natural assumption was that it did. I think this is actually good, Do not repeat.
- Variable and function names do not conform to `contributions.md`.
- If you set `LOG=1` and `TESTMODE=1` then you enter playback mode but do not log to a new database. Makes sense but simply noting that someone can attempt to setup the INI file like this.

I've let @jsr38 decide if he'd like to update the proxy contributions to follow the conventions we started following but since Proxy is and old contribution much of the code already written prior to this PR doesn't follow this either.

Issue #156, opened 2017-04-15
Investigate Updated Proxy To Twisted Module Use (Event Driven Proxy)


kb1lqd:
## Summary
Twisted is an event driven networking module for Python that also properly implements Serial support. This is advantageous due to the ability to move proxy from a polling based system to an event based system.

## Problem Explanation
Per comments from #138:
* https://github.com/FaradayRF/Faraday-Software/issues/138#issuecomment-292768375
* Regarding Async serial
  * "Twisted uses PySerial but it also properly implements asynchronous reads and writes on top of it using its own internal asynchronous I/O functions."

#### Software 
Standard Faraday software.

#### Hardware
Standard Faraday Hardware

# Supporting Information
See #138

Issue #138, opened 2017-03-29
proxy.py is not optimized for CPU consumption


ghost:
## Summary
When used with an embedded computer like a Raspberry Pi Zero wireless, the process consumes 95% of the CPU.

EDIT @kb1lqd : Created PR #157 

## Problem Explanation
The CPU load is 95% even though there is no radio traffic. The sleep() finctions are generally 1 ms, which may be too low. I changed them to 100 ms but the CPU was still consuming 85% of the CPU.

### Environment
Ubuntu Workstation on home network, connected to a Raspberry Pi Zero that will be remote near the antenna at altitude.

#### Software 
Latest commit:
https://github.com/FaradayRF/Faraday-Software/commit/19cb72217e6b6ee8e04527f8d969dc05a96ea936

#### Hardware
Raspberry Pi Zero running Debian on ARM6 hardware floating point:

Linux 4.4.50+ #970 Mon Feb 20 19:12:50 GMT 2017 armv6l GNU/Linux
Python 2.7.9

# Supporting Information

![top](https://cloud.githubusercontent.com/assets/24887005/24438825/d22391f8-140f-11e7-86d3-a1a18da1b42f.png)

reillyeon:
It would be nice to switch to waiting for the serial port to become readable instead of polling it for data. Unfortunately PySerial doesn't seem to support this cross-platform.

kb1lqc:
Thanks for the report @ObjectToolworks! Yes we need to address this at some point. I am happy you made this issue ticket so it's on our radar. Currently we are focusing on the computer based applications and #135 to help catch errors better and improve our speed of development.

I like what @reillyeon suggests and polling is not the best way to go but it was simple at the time. [MVP](https://en.wikipedia.org/wiki/Minimum_viable_product) as they say!

I found [this stackoverflow](https://stackoverflow.com/questions/20852060/pyserial-non-polling-read) thread which could be relevant though maybe not cross platform? @ObjectToolworks this could be implemented in proxy.ini however [line 340 of proxy.py](https://github.com/FaradayRF/Faraday-Software/blob/master/Proxy/proxy.py#L340) needs to be updated to allow a "None" type through... Could be a good starting point?

kb1lqd:
[...]

PR #157, opened 2017-04-15
Issue138 - Reduced CPU usages By Updating Proxy / UART Stack Efficiency

kb1lqd:
This PR captures the updates made from #138 which indicated that Proxy utilized an excessive amount of CPU usage during operations. Specifically a user was running proxy on a Rpi Zero but similar issues were noted on Rpi3 and a normal laptop to varying degrees.

Major change updates:
* Proxy no longer searches through ALL ports for TX and RX and instead identifies which ports are in use and only checks those ports (instead of all 256 possible)
* When new data for TX/RX is available the size of the queue is estimated an ALL known data items are retrieved/sent at once instead of a wasteful one-by-one pop off of a queue
* time.sleep() duration extended without known degradation (post updates as stated above)
* Proxy now creates a thread per unit connected and those threads only deal with a single unit each

TravisCI tests indicate passing.

Future Updates

Per the discussion on #138 but limited scope of this issue ticket I created a new enhancement issue ticket to capture the investigation and possible move to event driven proxy on #156 which introduces the Python Twisted module to eliminate polling operations in Proxy.


Proxy/proxy.py
```diff
@@ -459,36 +461,18 @@ def main():
...
-            time.sleep(1)
-
-    t = threading.Thread(target=uart_worker, args=(unitDict, getDicts, units, log))
-    threads.append(t)
-    t.start()
+    for key, values in units.iteritems():
+        unitDict[str(values["callsign"] + "-" + values["nodeid"])] = layer_4_service.faraday_uart_object(str(values["com"]), int(values["baudrate"]), int(values["timeout"]))
+
+    for key in unitDict:
+        print "Starting Thread For Unit:", key
```
kb1lqc: Please use logging!

Proxy/proxy.py
```diff
@@ -52,64 +52,66 @@ def uart_worker(modem, getDicts, units, log):
     """
     logger.info('Starting uart_worker thread')
 
+    unitcallid = str(units[units.keys()[0]]['callsign']) + '-' + str(units[units.keys()[0]]['nodeid'])
```
kb1lqc: This works but why not just grab station from `tempdict` key?

Proxy/faraday_uart_stack/layer_4_service.py
```diff
@@ -139,8 +139,13 @@ def RxPortHasItem(self, service_number):
         try:
             return not self.receive_parsed_queue_dict[service_number].empty()
         except:
-            self.receive_service_queue_open(service_number, self.QUEUE_SIZE_DEFAULT)
-            pass
+            return None
+
+    def RxPortListOpen(self):
+        return self.receive_parsed_queue_dict.keys()
```
kb1lqc, line 151: Where do you actually build the `receive_parsed_queue_dict` entries?

Proxy/faraday_uart_stack/layer_2_protocol.py
```diff
@@ -483,16 +484,18 @@ def GET(self):
...
-                data = self.receiver_class.rx_packet_queue.get()
-                try:
-                    unpacked_datalink = self.datalink_packet_struct.unpack(data)
-                    #Place datalink payload into payload queue
-                    self.rx_data_payload_queue.put(unpacked_datalink[3])
-                except:
-                    print "FAIL"
-                    pass  #print "Failed parsing" !!!!!FIX!!!!!
+                #  Loop through all known queue data and receive
+                for i in range(0, self.receiver_class.rx_packet_queue.qsize()):
```
kb1lqc: This could just be `...range(self.receiver_class.rx_packet_queue.qsize())`

Proxy/faraday_uart_stack/layer_2_protocol.py
```diff
@@ -483,16 +484,18 @@ def GET(self):
...
+                    data = self.receiver_class.rx_packet_queue.get()
+                    try:
+                        unpacked_datalink = self.datalink_packet_struct.unpack(data)
+                        #Place datalink payload into payload queue
+                        self.rx_data_payload_queue.put(unpacked_datalink[3])
+                    except:
+                        print "FAIL"
```
kb1lqc: Don't mean to be picky but lets just update this to use logging and remove line 495 inline commented out print statement too.

Proxy/faraday_uart_stack/layer_2_protocol.py
```diff
@@ -184,14 +184,15 @@ def run(self):
...
 
-                #New data available, retrieve single data "packet"
-                packet = self.insert_data_class.tx_packet_queue.get()
-
-                #Transmit
-                for i in range(0, len(packet), 1):
-                    self.output_channel.put(packet[i])
-                    self.serial_physical_obj.serial_physical_obj.send_byte(packet[i])
+                    #Transmit
+                    for i in range(0, len(packet), 1):
```
kb1lqc: Might as well update to `range(len(packet))`:

kb1lqc (changes requested):
Simple updates! Otherwise it appears good.

Proxy/proxy.py
kb1lqd: You're right, I replaced with modem['unit']

Proxy/proxy.py
kb1lqd: Update: logger.info('Starting Thread For Unit: ' + str(key))

Proxy/faraday_uart_stack/layer_4_service.py
kb1lqd: It is built in the initialization of the class object:  
image

Entries are added when "opening" a port" on line 144

image


Which proxy.py uses on line 66:

image

Proxy/faraday_uart_stack/layer_2_protocol.py
kb1lqd: Updated

Proxy/faraday_uart_stack/layer_2_protocol.py
kb1lqd: Updated, I tend to like explicitly showing intention but that's just me...

Proxy/faraday_uart_stack/layer_2_protocol.py
kb1lqd: Ok updated

Proxy/faraday_uart_stack/layer_4_service.py
kb1lqc: OK so when proxy performs `RXPortListOpen()` within that function it ends up running `receive_service_queue_open()` if an exception occurs. Cool.

kb1lqc:
OK not sure why but Travic-CI isn't starting builds reliably today... I checked this branch with a local run on `pytest` and it passed so I expect this to pass once Travic-CI builds.

The only thing I'd like to see is the commented out code simply deleted. We have a chance to do that now so why not. It's cleaner. Otherwise this looks good. I tested with telemetry and aprs which worked fine. CPU usage on my Surface Pro 4 was < 1% 🥇 and testmode worked fine. I'll take a leap of faith and say that Hermes and others will work with that too.

kb1lqd:
Just pushed a quick removal of extra code that was commented.

kb1lqc (approved):

PR #160, opened 2017-04-16
Move the Faraday UART library into the faraday module

reillyeon:
This moves Proxy/faraday_uart_stack to faraday/uart and updates proxy.py to import it from there.


kb1lqc:
@reillyeon Maybe I am missing something but I thought these updates were aimed to get Faraday importing like a module. Why the move towards importing with `os`? is this temporary until the moves are completed?

reillyeon:
The next step is to package the `faraday` module so that it can be installed into the system's Python path with setuptools/pip. Then the code to modify the Python path to include a local directory will be unnecessary. As a bonus we will be able to install apps like the proxy into the user's path as well.

Future:
```
# git clone https://github.com/FaradayRF/Faraday-Software.git
# pip install --editable Faraday-Software
# faraday-proxy
```

Or, for users who don't need to have a checkout of the source we can also upload the built package to PyPi and it can be installed with `pip install faradayrf`.

kb1lqc:
OK @reillyeon I messed up hardcore on this one. Sorry. First off I thought I was committing to a local branch of your branch but apparently I was committing to your repositories branch. Secondly this PR isn't doing well with the changes to the uart protocol files in the latest PR's that you ended up moving in this PR any suggestion for how to do this well?

I'm wondering if starting new is better here... sorry.

reillyeon:
Closing this pull request. I will rebase and open a new one.

PR #161, opened 2017-04-16
Move the Hermes library into the faraday module

reillyeon:
This moves Applications/Hermes/hermesmessage.py and hermesobject.py into the faraday library as their own "hermes" module. The Hermes Flask server is updated to import them from there.


kb1lqc:
@reillyeon does this need the latest master pulled in to avoid the merge mess we saw with `proxy`? I don't think it does since it doesn't touch the uart code at all.

reillyeon:
This patch should apply cleanly.

kb1lqd:
I confirmed Hermes still operates as intended with two units manually testing.

@reillyeon This move is basically removing the "object" and IO functions of Hermes and putting them in the faraday module folder as a "tool" for the hermes application to utilize correct? I see the benefit of doing this and I'd like to make sure @kb1lqc agrees. This is similar to Proxy utilizing proxyio tools but specific to messaging applications. 

When I add ARQ and other functionality I suspect it'd be best to include those updates into this module folder for import into a new Hermes application.

Has my approval and only waiting for @kb1lqc to OK the architecture move.

kb1lqd (approved):
Hermes script updates and folder moves are consistent with intended Pull Request description and intent. I verified Hermes operates nominally through manual testing.

reillyeon:
Eventually the applications as well will move into the faraday module and the scripts will be shortened to simple autogenerated wrappers. This is how Python's setuptools supports packages that provide command line applications. It has the benefit that tests will be able to initialize any application with simply Python code.

kb1lqc:
Yes @kb1lqd you are correct and we already did this for Proxy. However since you had updated the UART code and @reillyeon had moved it we ended up seeing some issues with GIT. This will however be much cleaner of a merge!

Thanks for the help and testing. I'll merge it.

kb1lqc (commented):
Looks good!

PR #164, opened 2017-04-17
Move the Faraday UART library into the faraday module

reillyeon:
This moves Proxy/faraday_uart_stack to faraday/uart and updates proxy.py to import it from there.


kb1lqc (approved):
Checked with hardware running Proxy, Telemetry, DeviceConfig, and Simpleconfig. All seem to work just fine! LoggingConfig.ini uses old naming conventions just like most of Proxy but that's grandfathered code.

PR #166, opened 2017-04-19
SimpleUI Application

kb1lqc:
I created a simple user interface. No frills. It's pure Python Flask + Javascript/HTML. There's a decent amount of new code but it all boils down to the UI and whether it works or not.

- It assumes basic interfacing for simplicity in development... i.e. uses a configuration file to set callsigns instead of user input via web-browser as that adds a bit of code that overall isn't the mission of this application.
- ADCs are scaled to volts at the ADC pin in javascript for the UI but saved in the database as bits for posterity. I did not indicate units on the UI. Let me know if they really should be added.

@reillyeon I'd appreciate a quick review. I don't think we need a super in-depth review here for Javascript/HTML but I did set up JSHint and htmllint locally with node.js to check to what I believe the JQuery project requires as we aim to mimic in CONTRIBUTIONS.MD. We do not currently have any automated node.js testing so I just left the configuration documents in for local installs.

@kb1lqd I don't think you need to dive into the review if you don't want to so I'll base my merge off of Reilly's comments unless say otherwise.

The readme.md shows the basic configuration (minimal) and even a gif of the UI working. This is sufficient to perform a basic Amateur Radio High Altitude Balloon Mission as well as general telemetry/command interfacing needs. There are some known issues still present as described in #145 that I believe should not hold up this PR as they could be firmware issues unless proven not to be.

SimpleUI


Applications/SimpleUI/simpleui.ini
```diff
@@ -0,0 +1,19 @@
+[FLASK]
```
kb1lqd: You committed your personal simpleui.ini when is should be added to GitIgnore.

kb1lqd (changes requested):
Reviewing gpio allocations shows the need to change PORT 3 GPIO allocations since they are backwards from firmware:

https://github.com/kb1lqc/Faraday-Software/blob/simple_web_ui/faraday/proxyio/gpioallocations.py#L11

**Firmware Allocations**
https://github.com/FaradayRF/Faraday-Firmware/blob/swissue153/REVA_Faraday.h#L65
image


**Software Allocations**
https://github.com/kb1lqc/Faraday-Software/blob/simple_web_ui/faraday/proxyio/gpioallocations.py#L11

image

Applications/SimpleUI/static/simpleui.js
```diff
@@ -0,0 +1,103 @@
...
+                    iostate = ("000000000" + data[0].IOSTATE.toString(2)).substr(-8).split("");
+                    iostate.reverse();
+
+                    // Convert RFSTATE data to an array of values
+                    var rfstate = [];
+                    rfstate = ("000000000" + data[0].RFSTATE.toString(2)).substr(-8).split("");
+                    rfstate.reverse();
+
+                    // Populate GPIO data in IO section
+                    $(".GPIO0").text(gpio[0]);
```
kb1lqd, line 65: (With updated firmware per my firmware issue ticket)

When I command GPIO 7 I see GPIO 0 turn on:
image

Reviewing this in code composer shows that I properly bit shifted GPIO 7 into the MSB location that matches the excel document:

image

image

@kb1lqc please review you're code here to make sure that you are properly allocating the bitfield.

kb1lqd: Checking GPIO 4 shows I properly bitshift but you show GPIO 3 as changed:

image

image

kb1lqd: This actually makes sense because I send the telemetry down in little endian format but my parsing forced big endian:
https://github.com/kb1lqc/Faraday-Software/blob/simple_web_ui/faraday/proxyio/telemetryparser.py#L18

Swapping `<` for `>` didn't seem to fix it but it's a good clue.
could javascript also be forcing an endianess?

@Bryce please verify and see if you can fix, this is something trivial. If it's easier for me to just swap my bitfield around on Faraday let me know.

kb1lqd (changes requested):
Added review comments on GPIO bitfield endianess.

Applications/SimpleUI/simpleui.ini
kb1lqc: Removed

Applications/SimpleUI/static/simpleui.js
kb1lqc: Using the [new firmware](https://github.com/FaradayRF/Faraday-Firmware/issues/72#issuecomment-295424504) I can clearly see the following when I print out the GPIO state in javascript from SimpleUI:

GPIO7 HIGH
image

GPIO0 HIGH
image

GPIO1 HIGH
image


Given the documentation showing the following GPIO State packet I believe this is simple reversed in the Javascript array:
image

However... Measuring with a multimeter to confirm GPIO output state...

# Commanding GPIOs
It appears commanding is messed up. Here's what happens when I send commands to Faraday GPIO

GPIO0: D2
GPIO1: D1
GPIO2: D0
GPIO3: D3
GPIO4: D4
GPIO5: D5
GPIO6: D6
GPIO7: D7
GPIO8: D8

_Clearly Commanding GPIO0 through GPIO2 is incorrect_

# GPIO Telemetry
D0: GPIO7
D1: GPIO6
D2: GPIO5
D3: GPIO4
D4: GPIO3
D5: GPIO2
D6: GPIO1
D7: GPIO7
D8: GPIO8

_Clearly GPIO0 never turns on and D0-D6 telemetry is reversed on SimpleUI_

kb1lqc: After reversing GPIO in javascript with `gpio.reverse();` I obtain the following:

D0: GPIO0
D1: GPIO1
D2: GPIO2
D3: GPIO3
D4: GPIO4
D5: GPIO5
D6: GPIO6
D7: GPIO7
D8: GPIO8

AWESOME @kb1lqd ! Telemetry now corresponds to MSP430 pins (D0-D8 are silkscreen designators on the PCBA). So we've fixed 

- The telemetry coming down UART is now correctly indicating GPIO values
- SimpleUI correctly outputs GPIO values to the user

Not sure if we really need to care about endianness at the moment. My Javascript code to create the array performs an into the binary string + shift already which actually might be the source of the flip. Regardless it's working right now and we have one issue in the way:

-[ ] GPIO0-GPIO2 commands are flipped

kb1lqc: @kb1lqd so it appears as though from a commanding standpoint either the firmware is flipped or [GPIO allocations](https://github.com/kb1lqc/Faraday-Software/blob/simple_web_ui/faraday/proxyio/gpioallocations.py#L11) are flipped. Thoughts?

It appears as though GPIO allocations are ordered correctly so I would suspect commanding in firmware?

reillyeon (commented):
This change looks good to me though I defer to @kb1lqd for final approval because he is more familiar with the GPIO assignments which appear to be a sticking point.

There are a couple of things I would like to see eventually happen in this code:

* Build more of the UI and form handling logic with code so that it is less repetitive. Copy-pasted code tends to lead to errors.
* Use Grunt/npm to pull in our dependencies like Bootstrap and jQuery during the build process so that we don't have to check them into git.

kb1lqc:
@reillyeon agreed and there was no real long-term thought put into this. RHIT is launching a balloon in a week or two and needed a user interface. This is no-frills. Ideally I'd like to use OpenMCT as it really seems to hand "missions" and situational awareness well. However that's really aimed at telemetry/command and being internet based I'd like to see another dashboard or similar project be utilized.

kb1lqd:
Firmware GPIO allocations match hardware:
image


The Python GPIO allocations do NOT match hardware in GPIO 0-2 which is shown in the swapped behavior during commanding. Please swap GPIO 0 with BIT 2 and GPIO 2 with BIT 0 as shown in the image below.

Update: https://github.com/kb1lqc/Faraday-Software/blob/simple_web_ui/faraday/proxyio/gpioallocations.py#L11

as shown: 
image

kb1lqd (approved):

kb1lqd:
I pushed the GPIO swap to fix the last GPIO commanding issue. All looks good. @kb1lqc

Issue #168, opened 2017-04-19
Create Proxy Log File Repository


kb1lqd:
## Summary
With the merging of #152 proxy now has the ability to "replay" prior recorded data. This is especially useful for developers without Faraday hardware or when testing specific situations. A repository on GitHub should be created to allow the community access to known good logs for playback.

## Problem Explanation
Create a log folder (and update contributions guide) for proxy playback recordings. This should include known good logs of things such as:
* General single unit local telemetry
* Multiple local unit telemetry
* Local and remote unit telemetry

The repository should clearly identify what the log file contains and is inteded to be used for.

#### Software 
N/A
#### Hardware
N/A

# Supporting Information

kb1lqc:
Closing as we have completed this!

PR #169, opened 2017-04-21
Added initial proxy log repository folder

kb1lqc:
As we aim to get people the ability to test Faraday Software out without the need to purchasing any hardware, we also want to enable open source contributions from anyone whether they own hardware or not. This is one of many log files that will help us round out the ability to run Faraday without any hardware.

The log is described in the text file and will enabled the following programs to be run without hardware:
- Proxy
- Telemetry
- APRS
- SimpleUI

## APRS
While I'm probably asking for people to run this and use KB1LQC-1/2 and essentially make it unusable for me on APRS... if that's the case I'll change out the log file but I will also likely need to avoid using these station IDs in the future on APRS unless we figure out how to indicate otherwise to APRS-IS.

The stations are static and located here on aprs.fi with KB1LQC-1 in Grand Park and KB1LQC-2 in the lower right corner with a green star symbol.

image

Commit Message:
New folder proxy/examplelogs contains a single log file created for
testing non-hardware use of Faraday software in regards to telemetry. A
new files called logs.txt was created to help document these files and
should be updated as new logs are added. This helps address Issue #168.


proxy/examplelogs/logs.txt
```diff
@@ -0,0 +1,6 @@
+This text document helps define each example log file. These files help users demo as well as test and develop with FaradayRF software without the need for hardware.
+---
+
+Log: 4-20-2017_1.db
```
reillyeon: I recommend standardizing on YYYY-MM-DD-##.db as it sorts better alphabetically.

reillyeon (changes requested):

proxy/examplelogs/logs.txt
kb1lqc: Ohh good call.

reillyeon (approved):

kb1lqc:
Thanks @reillyeon! I just updated the readme.md which I realized had the old filename in it.

PR #172, opened 2017-04-24
Move proxy/examplelogs to Proxy/examplelogs

reillyeon:
If you're on a case-insensitive filesystem you won't notice the difference but git (and correct operating systems) consider "proxy" and "Proxy" to be different directories.


kb1lqc (approved):
Good catch! I need to clean up my local git checkout to fix this on my end too.

Issue #178, opened 2017-04-26
Make Proxy a Python Module


kb1lqc:
## Summary
We aim to simplify Faraday Software as it needs to and it's currently a pain to use as well as test. Moving Proxy to a module will help us automate testing as well as drastically improve how we interact with the program.

## Problem Explanation
Python modules can be [executed as a script](https://docs.python.org/2/tutorial/modules.html#executing-modules-as-scripts). This is how we should have Proxy start up and it's nearly there! Additionally, we should avoid using globals in Flask apps and attach the configuration values to the [Flask Object](http://flask.pocoo.org/docs/0.12/config/#configuration-basics).

Tasks:
- Make Proxy an executable module
- Avoid Flask globals

#### Software 
Master branch of Faraday-Software as of 4-25-2017

#### Hardware
Any hardware and even software Proxy "Test Mode" will benefit from this.

# Supporting Information
Suggestions per @reillyeon:

![image](https://cloud.githubusercontent.com/assets/331540/25420713/45917f6a-2a0e-11e7-944a-cbfaf722e0c2.png)

![image](https://cloud.githubusercontent.com/assets/331540/25420720/4dc1b5e2-2a0e-11e7-9e39-5ab2a042b6fa.png)

kb1lqc:
Implemented in #181. Complete!

kb1lqc:
Just realized @reillyeon we never checked for globals in the flask application as this ticket suggested.

PR #181, opened 2017-04-27
Convert Proxy to a Module

kb1lqc:
Basic implementation of moving proxy.py with actual Proxy code into `faraday/proxy.py` and using a simplified `Proxy.proxy.py` to import proxy with `from faraday import proxy`. This allows us to simply call the module as an executable with `proxy.main()` utilizing the `if __name__ == '__main__':` code.

Manually tested that normal UART, logging, and test modes work!

This addresses Issue #178


reillyeon (approved):
I'm going to trust you didn't change anything about proxy.py other than the imports at the top. 😄

kb1lqc:
@reillyeon haha yeah that's all I did in `/faraday/proxy.py`. All actual code changes performed on the last commit 1c57b8b2f3e56cd6fad3a52a238cd342c0e619c1 otherwise that file was only moved.

PR #192, opened 2017-04-30
Fix script permissions and #! lines

reillyeon:
Removed execute bit and #! lines from files in the faraday module and made sure that all the new wrapper scripts are executable.


kb1lqc:
Everything looks good and I had to check [this stack overflow](https://stackoverflow.com/questions/10516201/updating-file-permissions-only-in-git) to see how GitHub handles file permissions (new to me)!

Why do we need to remove the shebang #! line?

el-iso:
@kb1lqc I believe its because most users don't need to be running files in the faraday module directly. Where as the wrapper scripts are the actual "applications" that the user is intended to run. Not including the shebang just means that if someone does want to run the faraday module scripts they will need to use "python <script name>"

kb1lqc:
@el-iso 

homer_bushes

kb1lqc:
So to be clear the intention is to make the application wrapper scripts only executable with `python <filename>`?

el-iso:
@kb1lqc actually I think its the opposite. If you look at the changes @reillyeon  made, the files in /Applications still have the shebang, but the files in /faraday (the faraday module) do not. So the faraday module scripts (not the application wrappers) would require "python <filename>" to be executed.

kb1lqc:
Oh I am oblivious. That makes so much sense. Thanks for explaining the rational!

kb1lqc (approved):
With some help from @el-iso I now realize the rational behind this. Looks good!

Issue #132, opened 2017-03-14
Proxy Fails To Handle Incorrect Unit Header (count)


kb1lqd:
## Summary
When using multiple Faraday units (2) and leaving both the .INI headers [UNIT0] the proxy fails hard. Proxy should gracefully handle this and report the error to the user.

## Problem Explanation
When connecting proxy to multiple local units if the configuration (for 2 units) unit header is [UNIT0] and [UNIT0] instead of [UNIT0] and [UNIT1] the program crashes. Proxy should handle this gracefully and report the error to the use rather than crashing and leaving no debugging information. Otherwise this error should not be allowed to occure if it cannot be reported.

#### Software 
Software: Master

#### Hardware
SN: 12, 13

# Supporting Information

Below is an incorrect QTY=2 unit proxy.ini configuration.
![image](https://cloud.githubusercontent.com/assets/339144/23889009/17bb2e98-0848-11e7-9b6a-570b38e6b0f0.png)

The failure log below when run in a scripting IDE:

![image](https://cloud.githubusercontent.com/assets/339144/23889084/7d255056-0848-11e7-9659-76b46589516b.png)

The problem is resolved as expected when configured with the second unit as [UNIT1]

![image](https://cloud.githubusercontent.com/assets/339144/23889099/9130a212-0848-11e7-8dba-c12f85e59cb5.png)

kb1lqc:
Taking a stab at this I implemented proper try/except statement around the `callsign2COM()` function in `proxy.py`. This catches the `ConfigParser.NoSectionError` and exits python.

https://github.com/kb1lqc/Faraday-Software/blob/issue132/faraday/proxy.py#L584

**Proper operation**

```
C:\Users\Bryce\Documents\GitHub\faradayrf-kb1lqc\Proxy>python proxy.py
2017-04-30 09:41:09,815 - Proxy - INFO - Starting proxy server
2017-04-30 09:41:09,867 - Proxy - INFO - Starting Thread For Unit: KB1LQC-2
2017-04-30 09:41:09,868 - Proxy - INFO - Starting uart_worker thread
2017-04-30 09:41:09,868 - Proxy - INFO - Starting Thread For Unit: KB1LQC-1
2017-04-30 09:41:09,868 - Proxy - INFO - Starting uart_worker thread
```

**Exception Operation**

```
C:\Users\Bryce\Documents\GitHub\faradayrf-kb1lqc\Proxy>python proxy.py
[...]

PR #193, opened 2017-04-30
Added try/except statement for units config

kb1lqc:
When creating appropriate items based on proxy.ini configuration file
sections a new try/except statement catches the
ConfigParser.NoSectionsError. When the exception is caught it logs the
error and exits python.

This addresses and fixes issue #132.

### DIFF
--- a/faraday/proxy.py
+++ b/faraday/proxy.py
@@ -585,7 +585,7 @@ def main():
         units = callsign2COM()
     except ConfigParser.NoSectionError as e:
         logging.error(e)
-        sys.exit()
+        sys.exit(1) #  Status Code 1 = Config Error
 
     if testmode == 0:
         for key, values in units.iteritems():
--- a/faraday/proxy.py
+++ b/faraday/proxy.py
@@ -581,7 +581,12 @@ def main():
 
     # Associate serial ports with callsigns
     # global units
-    units = callsign2COM()
+    try:
+        units = callsign2COM()
+
+    except ConfigParser.NoSectionError as e:
+        logging.error(e)
+        sys.exit()
 
     if testmode == 0:
         for key, values in units.iteritems():
--- a/faraday/proxy.py
+++ b/faraday/proxy.py
@@ -585,7 +585,7 @@ def main():
         units = callsign2COM()
     except ConfigParser.NoSectionError as e:
         logging.error(e)
-        sys.exit(1) #  Status Code 1 = Config Error
+        sys.exit(1)  # Status Code 1 = Config Error
 
     if testmode == 0:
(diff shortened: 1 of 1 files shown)


el-iso (commented):
This code looks good to me. Provided the only issue was the "NoSectionError" described in the original issue ticket, this should make proxy fail more gracefully.

Will this make it obvious to the user what they did wrong? 

I'll let someone else give final approval since I am not very familiar with this particular issue.

faraday/proxy.py
```diff
@@ -581,7 +581,12 @@ def main():
 
     # Associate serial ports with callsigns
     # global units
-    units = callsign2COM()
+    try:
+        units = callsign2COM()
+
```
reillyeon: Style nitpick: Don't put a space between the `try` and `catch` blocks since they are logically connected, at least when the `try` block is only one line.

faraday/proxy.py
```diff
@@ -581,7 +581,12 @@ def main():
...
+    except ConfigParser.NoSectionError as e:
+        logging.error(e)
+        sys.exit()
```
reillyeon: On error Python should exit with a non-zero error code so when catching errors like this we should use `sys.exit(1)`. (Coming up with a scheme where 1 means configuration error, 2 means something else, etc. is entirely optional and more work than really necessary.)

reillyeon (changes requested):
If main() throws an exception the application will print it out and exit anyways. What is the purpose of this?

reillyeon:
Ignore that comment about why the try/catch is needed at all. I forgot to delete it before submitting the review.

faraday/proxy.py
kb1lqc: Fixed in https://github.com/FaradayRF/Faraday-Software/pull/193/commits/5db3383d59ad59ebe48530e1cf210c0b5c2a5334

faraday/proxy.py
kb1lqc: Updated in https://github.com/FaradayRF/Faraday-Software/pull/193/commits/49a24e73af1a7b220b169340dc2ada718ce64fae

Picking status code 1 = configuration error. Where do we think this should be documented? I'm thinking either in `contributions.md` or a separate `errors.md` type file?

reillyeon (approved):

kb1lqc:
@reillyeon pushed commit https://github.com/FaradayRF/Faraday-Software/pull/193/commits/3cb1b97ea4387e1bb8190f40f94335ffc95dc616 after you approved which fixed a PEP8 failure I let through. Letting Travic-CI catch up before merging. Thanks!

Issue #62, opened 2017-02-08
Proxy cannot connect to only one unit if configured for two units


jbemenheiser:
When I have two units in the the proxy.ini file, proxy won't start if only one unit is plugged in.

Using the following proxy configuration, when both units are plugged in proxy.py starts properly.
![image](https://cloud.githubusercontent.com/assets/25631614/22727735/4a0b5a7c-ed8f-11e6-831c-11b66b2bedec.png)

However, with only the unit plugged into COM4 proxy.py shows this error.
![image](https://cloud.githubusercontent.com/assets/25631614/22727613/aa9c6c56-ed8e-11e6-9747-7dbbdc45437b.png)

Still with only the unit plugged into COM4 and commenting out the first unit in proxy.ini, proxy.py was able to start for the one unit that was connected.
![image](https://cloud.githubusercontent.com/assets/25631614/22727443/bc9cc604-ed8d-11e6-842f-1b9627554ce4.png)

kb1lqd:
This should probably just softly fail that port and continue with a warning/error for the unconnected port.

jbemenheiser:
Yes, that would make it much easier for a user who has two units but is switching between one or two of them connected to the computer.

kb1lqc:
Master Commit branched: https://github.com/FaradayRF/Faraday-Software/commit/2f0599841f91e79822ea1d6ee011afc06d87f15c

Testing this out with the following configuration:
```
[FLASK]
HOST=127.0.0.1
PORT=8000

[PROXY]
UNITS=2

[UNIT0]
CALLSIGN = KB1LQC
NODEID = 1
COM = COM42
BAUDRATE = 115200
TIMEOUT = 5

[UNIT1]
CALLSIGN = KB1LQC
NODEID = 2
COM = COM13
BAUDRATE = 115200
TIMEOUT = 5
```

Proxy appears to start up fine with this output when I print the `units` parameter variable in `uart_worker()` on line 44 of proxy.
```
2017-02-23 00:47:30,930 - Proxy - INFO - Starting proxy server
2017-02-23 00:47:30,996 - Proxy - INFO - Connected to Faraday
2017-02-23 00:47:31,002 - Proxy - INFO - Starting uart_worker thread
[...]

PR #195, opened 2017-04-30
Issue62 Fix: USB Cable Connection Exceptions

kb1lqc:
This PR fixes the complaints in #62 which essentially deal with the USB connection. Both scenarios are now caught and handled.

- Removing USB during `Proxy` operation
```
C:\Python27\python.exe C:/Users/Bryce/Documents/GitHub/faradayrf-kb1lqc/Proxy/proxy.py
2017-04-30 11:10:29,854 - Proxy - INFO - Starting proxy server
2017-04-30 11:10:29,903 - Proxy - INFO - Starting Thread For Unit: KB1LQC-2
2017-04-30 11:10:29,903 - Proxy - INFO - Starting uart_worker thread
2017-04-30 11:10:29,904 - Proxy - INFO - Starting Thread For Unit: KB1LQC-1
2017-04-30 11:10:29,904 - Proxy - INFO - Starting uart_worker thread
2017-04-30 11:10:31,104 - root - ERROR - USB cable likely disconnected!
2017-04-30 11:10:31,104 - root - ERROR - call to ClearCommError failed
```

- Starting `Proxy` with any USB connection missing
```
C:\Python27\python.exe C:/Users/Bryce/Documents/GitHub/faradayrf-kb1lqc/Proxy/proxy.py
2017-04-30 11:04:46,812 - Proxy - INFO - Starting proxy server
2017-04-30 11:04:46,838 - root - ERROR - Check USB cable connection!
2017-04-30 11:04:46,838 - root - ERROR - could not open port 'COM42': WindowsError(2, 'The system cannot find the file specified.')
```

## 7-6-2017 UARTstack Comment

@reillyeon it appears @kb1lqd doesn't want to remove the [UART service layer 2 object](https://github.com/FaradayRF/Faraday-Software/pull/195#discussion_r125560959). He does make a good point that the UART code is SO old that we really should consider completely rewriting it one day. There are a ton of brute force timer sleeps and such in it that I have no doubt will become a bottleneck one day and force us to reconsider the architecture of the class itself. This will probably happen in the near future with data applications on the horizon. **Therefore we're punting this concern to a later issue ticket**


kb1lqc:
I know I used a different method of sys.exit() in this fix. Instead of importing all of sys I just imported exit() from sys. Not sure if I should keep it consistent. This seemed like less of a footprint.

reillyeon:
I prefer not importing individual functions from a module because when you see `sys.exit` you know what it means whereas `exit` could mean anything. Also, `sys.exit(1)` since these are failures. To clarify my earlier comment: I don't believe it is worthwhile to define failure codes other than 0 for success, 1 for error.

kb1lqc:
@reillyeon fixed and yes that actually makes more sense. Simply using exit code of 1 for anything incorrectly exiting.

faraday/uart/layer_2_protocol.py
```diff
@@ -15,14 +15,20 @@
 import time
 import struct
 import logging
+import sys
 
 # Get configured logger
 logger = logging.getLogger('UARTStack')
 
 
 class layer_2_object(object):
```
reillyeon, line 24: Can we just remove this class entirely? It looks like it doesn't serve any purpose than to hold a layer_2_protocol object.

faraday/uart/layer_2_protocol.py
```diff
@@ -67,9 +73,14 @@ def run(self):
...
-                    rx_buffer_inwaiting = self.ser.inWaiting()
-                    self.serial_rx_queue.put(self.ser.read(rx_buffer_inwaiting))
+                try:
+                    if self.ser.inWaiting() > 0:
+                        rx_buffer_inwaiting = self.ser.inWaiting()
+                        self.serial_rx_queue.put(self.ser.read(rx_buffer_inwaiting))
+                except serial.SerialException as e:
+                    logging.error("USB cable likely disconnected!")
+                    logging.error(e)
+                    sys.exit(1)  # Sys.exit(1) is an error
```
reillyeon: Why not call `self.abort()` here instead? I'm concerned that adding `sys.exit` calls too deeply layered in the code will make it difficult for us to decide later what errors are fatal or not.

faraday/uart/layer_2_protocol.py
kb1lqc: I beleive @kb1lqd implemented it as an abstraction. I can dig into it too.

faraday/uart/layer_2_protocol.py
el-iso: @reillyeon is abort() from "os"? Are you suggesting that it would be better because it triggers a core dump? Just trying to educate myself on this subject....

reillyeon: `abort()` is a method on this class that sets `self.enabled` to `false` which will cause this loop to terminate and the thread to exit.

el-iso: Oh, that was dumb of me. I didn't even think to look for the implementation here... Thanks

faraday/uart/layer_2_protocol.py
```diff
@@ -15,19 +15,28 @@
...
 # Get configured logger
 logger = logging.getLogger('UARTStack')
 
 
 class layer_2_object(object):
     def __init__(self, port, baud, timeout):
-        self.serial_physical_obj = layer_2_protocol(port, baud, timeout)
+        # Start up serial object and if it fails exit the program
+        try:
+            self.serial_physical_obj = layer_2_protocol(port, baud, timeout)
```
reillyeon: Move this try/catch block into the layer_2_protocol constructor so it is clear what SerialException is being caught (the one that might be thrown by the serial.Serial constructor).

reillyeon (approved):
This looks good to me with one nitpick about where to place a try/catch block.

kb1lqc:
Updated this branch with the latest master code with a merge. All seems well. Used `pip install -e .` to install and edit the later updates to logging and error handling.

In addition to ensuring correct exceptions when USB cable is removed I also updated the logging use in the UART services. Below is an example of a UART communications error that results in an unsuccessful struct.unpack(). I now use appropriate warnings level and log the data in text format.

image

faraday/uart/layer_2_protocol.py
kb1lqc: Just pushed an update addressing this comment. Is this what you wanted?

reillyeon: Yes.

faraday/uart/layer_2_protocol.py
kb1lqc: @reillyeon it appears @kb1lqd used this in [layer 2 service](https://github.com/kb1lqc/Faraday-Software/blob/fec6feae17b3e843459d6582d7040354bba2c51a/faraday/uart/layer_2_service.py#L26) I think it probably can be removed if this line is updated. Am I correct Brent?

kb1lqc:
This PR is ready to merge pending @kb1lqd thoughts on removing the layer 2 object

kb1lqd:
Testing indicates some questions.

I used this PR to write some quick batch scripts: [faraday-bat.zip](https://github.com/FaradayRF/Faraday-Software/files/1125969/faraday-bat.zip)

# Unit disconnections:

2 units were connected and I connected telemetry to the unit-1 (COM8). Disconnecting unit-2 (COM112) showed the intended error message and telemetry continued to be saved from the remaining connected unit.

image

# Starting proxy with one unit disconnected
I kept proxy with option `--number 2` but disconnected the COM112 unit and started proxy. Proxy refuses to start:

image

@kb1lqc I thought that the intended operation was that proxy WILL start if one of the units is missing? What is the intended operation?

etc/faraday/loggingConfig.ini
```diff
@@ -36,7 +36,7 @@ qualname=SimpleConfig
 propagate=0
 
 [logger_UARTStack]
-level=INFO
+level=WARNING
```
kb1lqd, line 39: OK

faraday/uart/layer_2_protocol.py
kb1lqd: I seem to only use it locally in layer_2_object() class. I agree that this may be a bit convoluted and unneeded but the protocol object itself certainly is used. I believe it's best to leave it here for now and have a seperate IT/PR to update the lower level UART code for cleanliness/optimization. We're going to need to likely do this anyways with the improvement of generic data operation on the near term horizon.

kb1lqd (changes requested):
Looks good but my only request is to clarify operation when proxy is started expecting multiple units but only a single unit is connected? I seem to experience unintended failure of proxy booting in this condition.

kb1lqc:
Ok @kb1lqd I'll look at it in the morning. Multiple units is something I didn't eve consider tbh. Good catc

kb1lqc:
@kb1lqd I commented in #62 with the additions that make multiple unit configurations work better and fail gracefully. I believe this is all you need to be OK with approving and merging? Please do so if you are OK with these changes!

kb1lqd:
@kb1lqc Thanks for the update, just to clarify: I still cannot start proxy with only one unit connected but proxy expecting 2 units.
image

I don't see the text failure that you implemented here: https://github.com/FaradayRF/Faraday-Software/pull/195/commits/37b0100b928f5be40554e554b7d8c6794b4e3db9#diff-a9d2cba55827323291e9d0c49fd6f720R818

All in all it seems to be working but this one case isn't handled. I'd expect proxy to connect to all available units connected and ignore those not connected but configured to connect to by error'ing on them but not blocking operation. Is this not the expected behavior?

I'm cool with approving this if we acknowledge that you EXPECT proxy to not operate when:
- Config specifies 2 units connected
- Only a single unit connected when proxy boots

@kb1lqc Is https://github.com/FaradayRF/Faraday-Software/pull/235 suppose to fix my question above?

faraday/proxy.py
```diff
@@ -811,7 +811,11 @@ def main():
 
     if testmode == 0:
         for key, values in units.iteritems():
-            unitDict[str(values["callsign"] + "-" + values["nodeid"])] = layer_4_service.faraday_uart_object(str(values["com"]), int(values["baudrate"]), int(values["timeout"]))
+            try:
+                node = str(values["callsign"] + "-" + values["nodeid"])
+                unitDict[node] = layer_4_service.faraday_uart_object(str(values["com"]), int(values["baudrate"]), int(values["timeout"]))
+            except:
+                logger.error('Could not connect to {0} on {1}'.format(node, values["com"]))
```
kb1lqd, line 818: OK good. Seems to work.

kb1lqd (approved):
Latest commit seems to address my concerns.

PR #200, opened 2017-05-10
Fix Master to Run From Command Line

kb1lqc:
Master has been broken for a few days. Honestly not sure how it happened and we could use blame to figure out but not really worth the time. Likely some confusion on moving to packages.

**Problem**: You couldn't run Faraday Master Software from the command line by calling 

- `python Proxy/proxy.py`
- `python Applications/Telemetry/telemetry.py`
- `python Applications/APRS/aprs.py`
- `python Applications/SimpleUI/simpleui.py`

**Solution**: I added `os.path` imports to use the `__file__` path which uses the directory the module was imported from as the starting point. This would be the `faraday` folder in all current cases. This should also future proof us and make it easier for packaging modules as we're currently having a heck of time moving to them!

I did not bother update any other applications at this time in effort to get this right for packaging. One step at a time.

### DIFF
--- a/faraday/aprs.py
+++ b/faraday/aprs.py
@@ -14,16 +14,21 @@ import threading
 import ConfigParser
 import socket
 import requests
+import os
 from time import sleep
 
 # Start logging after importing modules
-logging.config.fileConfig('loggingConfig.ini')
+dir = os.path.dirname(__file__)
+logConfig = os.path.join(dir,"..","Applications","APRS","loggingConfig.ini")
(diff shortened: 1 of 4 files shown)


kb1lqc:
Hmm this PR seems irrelevant with the new packaging. Might as well just push towards completion with packaging.

kb1lqc:
As of PR #219 all applications now start from command line. This PR is irrelevant and the methodology is no longer used. Closing.

Issue #191, opened 2017-04-29
Use Python Setuptools to Install Faraday Software


kb1lqc:
## Summary
Let's face it, installing and running Faraday software is a pain right now. Thanks to #178, #182, #184, #185, and #187 we can finally move towards using `setuptools`. The end result will be that Faraday Software can be installed with `python setup.py <some_command> <options>` and run by just invoking the name of the software. Additionally, released Faraday Software would be registered with Pypi allowing installation from pip instead of relying on GitHub (unless you want the latest version).

## Problem Explanation
'nuff said.

#### Software 
Latest Master version as of 4-29-2017 and the general Python environment.

#### Hardware
N/A

# Supporting Information
[setuptools](https://pythonhosted.org/an_example_pypi_project/setuptools.html)
Official [pypi distribution guide](https://packaging.python.org/distributing/#setup-py)

kb1lqc:
Hey that's cool! haven't tested it from the repo but apparently I followed the pypi steps necessary to get Faraday into the package list! @kb1lqd @reillyeon 

Actual Package (not playing well with markup readme.md): https://pypi.python.org/pypi/faraday/0.0.1a1

![image](https://cloud.githubusercontent.com/assets/331540/25568964/2aee0c2a-2dc2-11e7-8e34-9eb7b22fa6f3.png)

kb1lqc:
Yup work to do but neat to be able to even get a response running `pip install faraday` 🍾 

![image](https://cloud.githubusercontent.com/assets/331540/25568997/a4203546-2dc2-11e7-837e-7854ee51c11e.png)

el-iso:
@kb1lqc  I got the same result. I'm glad you posted. I was trying to figure out what I did wrong! haha

kb1lqc:
Update! I used the `pip --log` parameter to generate a log file of my pip failure while using a virtualenv. Interesting. I have the following gist:

[pipfaraday.txt](https://gist.github.com/kb1lqc/6686b072766de311baf6bc42f3390086)

For some reason [pypi is defaulting to V0.0.1](https://gist.github.com/kb1lqc/6686b072766de311baf6bc42f3390086#file-pipfaraday-txt-L15) instead of the newer versions... possibly because it's the only tar.gz file?

Also I fell victim to this [pip bug](https://github.com/pypa/pip/issues/2762) that I am not sure is actually solved...

kb1lqc:
[...]

PR #203, opened 2017-06-03
Faraday-Proxy PyPi Support

kb1lqc:
Related to Issue #191 

This PR is intended to get the Proxy application into the pip index. It also allows for "editable" mode installation with command line operation of proxy.

This PR will definitely need some love before merging but it's a good starting point for discussion

Operation

Navigate to the main folder and ```pip install -e .``` to install in editable mode. One can also follow the python packaging guides to create an uploadable distribution for the PyPi servers.

Once installed simply run "Faraday-Proxy" to run the proxy program from command line.

You will need to configure the proxy application by first seeing the options by running `faraday-proxy -h`

First you will need to create a new configure file by running `faraday-proxy --configure` then run a second Faraday-Proxy command with the appropriate options to set the callsign, node Id, and port (usually comx on windiws).

This is the best I got for remembering from memory try to figure it out the best you can with the help options and then asking me the rest. I don't have access to the software as I'm traveling internationally at this time without a laptop!


kb1lqd:
I cloned the branch successfully. Do note @kb1lqc that I could `pip install faraday` before cloning...

kb1lqd:
@kb1lqc your command 1`pip install -e .` failed:

image

kb1lqd:
It looks like I needed to `pip install setuptools` and now seems to have successfully installed using `pip install -e .`

image

kb1lqd:
Ahh ok `pip install faraday` doesn't seem to let using `faraday-proxy` from CMD but this `pip install -e .` does.

kb1lqd:
One annoying thing is that `Faraday-Proxy` doesn't autocomplete using TAB.

kb1lqd:
`faraday-proxy --configure` is unrecognized

image

@reillyeon any thoughts?

kb1lqd:
Oh no, I think this is messing with all my other faraday branches I have checked out... I can run applications and such from other branches after install this PR... I need to uninstall to run any other heritage branches?

EDIT:

I needed to `pip uninstall faraday` to run any other GITHUB branches in our heritage fashion.

el-iso:
Yes, I experienced a similar problem. I think the package installed via pip
takes precedence over some other stuff in the code. After I uninstalled the
package it fixed it

kb1lqc:
Yes please do not use pip install Faraday at this time. In learning PyPi I uploaded a non working package to the main PyPi site instead of test PyPi. That version us very old and won't work. You must install editable or from PyPi test server.

However the PyPi test server will often fail on a fresh install since not all required packages have test PyPi uploads with the latest versions so you need to install thise from normal PyPi instead of test.

https://stackoverflow.com/a/34561435

This is the latest test PyPi which should be the same version in my git PR
https://testpypi.python.org/pypi/faraday/0.0.1009.dev988

Use the stack overflow method above to install from the test PyPi if you want. However I still suggest editable mode!

kb1lqc:
@kb1lqd @el-iso woops! --init-config is what you want to create the initial config file, sorry about the confusion!

After that runs you'll have a config file with REPLACEME placeholders. Then run in one or more command line entries

Faraday-Proxy --callsign CALLSIGN --nodeid NODEID --port PORT

Make sure you remember port is com port so COM5 or similar on Windows and dev/ttyusb or similar on Linux and mac

kb1lqd:
@kb1lqc 

`pip install -e .` Looks to have successfully installed faraday:

image

I made sure to restart the command window just incase.

`faraday-proxy --init-config` fails!

image

It looks like it is not expecting arguments but getting them?

_**Also, I definitely could NOT run my RFDataport branch with this installed so I had to remove it again.**_

kb1lqc:
Hmm make sure you have the python module shutil though I think it comes with python anyways. That said weird, it's complaining that I passed the initial argument which should be true or false...

kb1lqc:
Yup @kb1lqd looks like line 86 of proxy.py clearly shows no arguments yet like 215 passes them. Not sure how this happened. Try modifying line 215 to remove the argument passed. Installed in editable mode literally means you can edit the file and try again, no need to re install

kb1lqd:
@kb1lqc @reillyeon  I verified that Bryce's suggestion to remove line 215 argument in proxy.py worked and I can now run proxy as described:

image

kb1lqc:
Running the current commit https://github.com/FaradayRF/Faraday-Software/pull/203/commits/b6798b4e1bdf750798e0aa60226d41e26a411abe results in the following error if not launched from the main folder. This test was performed by moving up one directory.

```
Bryce@BRYCE-SP4 MINGW64 ~/Documents/GitHub
$ faraday-proxy
Traceback (most recent call last):
  File "C:\Python27\Scripts\faraday-proxy-script.py", line 11, in <module>
    load_entry_point('faraday', 'console_scripts', 'faraday-proxy')()
  File "C:\python27\lib\site-packages\pkg_resources\__init__.py", line 560, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "C:\python27\lib\site-packages\pkg_resources\__init__.py", line 2648, in load_entry_point
    return ep.load()
  File "C:\python27\lib\site-packages\pkg_resources\__init__.py", line 2302, in load
    return self.resolve()
  File "C:\python27\lib\site-packages\pkg_resources\__init__.py", line 2308, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "c:\users\bryce\documents\github\faradayrf-kb1lqc\faraday\proxy.py", line 216, in <module>
    configureProxy(args, proxyConfigPath)
  File "c:\users\bryce\documents\github\faradayrf-kb1lqc\faraday\proxy.py", line 173, in configureProxy
    config.set(unit, 'BAUDRATE', args.baudrate)
  File "C:\python27\lib\ConfigParser.py", line 396, in set
    raise NoSectionError(section)
ConfigParser.NoSectionError: No section: 'UNIT0'
```

It appears that the code for finding `proxy.ini` is wrong and defaulting to the current directory.

faraday/proxy.py
```diff
@@ -19,21 +19,218 @@
...
+userpath = os.path.join(os.path.expanduser('~'), '.faraday')
+path = ''
+
+for location in os.curdir, relpath1, relpath2, setuppath, userpath:
+    try:
+        logging.config.fileConfig(os.path.join(location, "loggingConfig.ini"))
+        path = location
+        break
+    except ConfigParser.NoSectionError:
+        pass
```
kb1lqc, line 44: Because I don't fail gracefully this code allows editable mode (or other installations) to fail if the file is never found. The code simply moves on with a blank path. Woops. This is related 
to IT #206

faraday/proxy.py
```diff
@@ -19,21 +19,218 @@
...
+
 logger = logging.getLogger('Proxy')
+logger.debug("PATH: " + path)
+
+#Create Proxy configuration file path
+proxyConfigPath = os.path.join(path, "proxy.ini")
+logger.debug('Proxy.ini PATH: ', proxyConfigPath)
```
kb1lqc: Woops, this is wrong. I need to use a `+` instead of comma to join the two items inside the logging.debug() code. Also, this is redundant and I should delete line 47

faraday/proxy.py
```diff
@@ -19,21 +19,218 @@
...
+parser.add_argument('--init-config', dest='init', action='store_true', help='Initialize Proxy configuration file')
+parser.add_argument('--callsign', help='Set Faraday callsign')
+parser.add_argument('--nodeid', type=int, help='Set Faraday node ID')
+parser.add_argument('--port', help='Set Faraday UART port')
+parser.add_argument('--baudrate', default='115200', help='Set Faraday UART baudrate')
+parser.add_argument('--timeout', type=int, default=5, help='Set Faraday UART timeout')
+parser.add_argument('--unit', type=int, default=0, help='Specify Faraday unit to configure')
+
+#Proxy options
+parser.add_argument('--number', type=int, default=0, help='Set number of Faraday radios to use')
```
kb1lqc, line 63: This may be dangerous and/or not work. Keep an eye on this one. Might need to default to 1?

faraday/proxy.py
```diff
@@ -19,21 +19,218 @@
...
+    '''
+
+    logger.info("Initializing Proxy Log File")
+    log = config.get("DATABASE", "filename")
+    logpath = os.path.join(os.path.expanduser('~'), '.faraday', 'lib', log)
+    os.remove(logpath)
+    logger.info("Log initialization complete")
+
+
+def saveProxyLog(name, config):
```
kb1lqc, line 114: May not work without pip install creating ~user/.faraday/lib

faraday/proxy.py
```diff
@@ -19,21 +19,218 @@
...
+    '''
+
+    log = config.get("DATABASE", "filename")
+    oldpath = os.path.join(os.path.expanduser('~'), '.faraday', 'lib', log)
+    newpath = os.path.join(os.path.expanduser('~'), '.faraday', 'lib', name)
+    shutil.move(oldpath, newpath)
+    sys.exit(0)
+
+
+def showProxyLogs():
```
kb1lqc, line 130: May not work without pip install creating ~user/.faraday/lib

kb1lqc: Not applicable, folder created if not already existing whenever proxy is run:

https://github.com/FaradayRF/Faraday-Software/pull/203/files#diff-a9d2cba55827323291e9d0c49fd6f720R629

faraday/proxy.py
kb1lqc: Not applicable, folder created if not already existing whenever proxy is run:

https://github.com/FaradayRF/Faraday-Software/pull/203/files#diff-a9d2cba55827323291e9d0c49fd6f720R629

requirements.txt
```diff
@@ -1,4 +1,5 @@
 appdirs==1.4.3
+argparse==1.4.0
```
kb1lqc: Shouldn't need this since python >2.6 has argparse in the stdlib

kb1lqc: Also should possibly add setuptools?

faraday/proxy.py
kb1lqc: This is actually OK. Defaulting to 0 basically keeps the 1 in the `proxy.ini` file until you specify something other than 0 or 1 (since 0 is ignored... you'd never have 0 units to startup. Maybe in testmode but that's getting in the weeds for now). Programming a number.

This does allow negative numbers and any number that is positive

kb1lqc: OK this is addressed in Issue #207

kb1lqd (approved):
I sat down with Bryce and we went through all of the code updates. I verified that I could install, configure, and start proxy from the command line. I also verified that telemetry and and APRS also worked (also commanding local and remote) using traditional methods of manually running the github scripts.

kb1lqc:
Uploaded to PyPi servers! @el-iso  @reillyeon @kb1lqd

TestPyPi: https://testpypi.python.org/pypi/faraday/0.0.1010.dev996
PyPi: https://pypi.python.org/pypi/faraday/0.0.1010.dev996

Issue #209, opened 2017-06-22
faraday-telemetry Should Be Updated To Support Multiple Units From Configuration


kb1lqc:
## Summary
Initially, faraday-telemetry does not properly read multiple faraday radio's connected via proxy into the program for data aggregation. This has to do with how the telemetry.ini file is read and telemetry_worker() threads are spun up.

## Problem Explanation
N/A

#### Software 
Faraday Software

#### Hardware
N/A

# Supporting Information
N/A

PR #211, opened 2017-06-22
Provide faraday-telemetry command line functionality

kb1lqc:
This PR provides the code necessary to:
* Install faraday-telemetry while performing a `pip install faraday`. Currently only usable in the editable local mode since at the time of the PR creation no new PyPi packages have been uploaded.
* Configure faraday-telemetry via command line
* Command certain functionality such as log saving and deletion via command line

Overall this PR is to get this code exposure. It needs to be cleaned up a tad but is 98% there. Some functionality has not been implemented but was captured in #209. Also, any package built from this PR will provide `faraday-proxy` and `faraday-telemetry` after the pip installation.


kb1lqd:
* Install worked as expected

* ISSUE: If I delete the config file in the telemetry folder and `--init-config` the file never appears again... I don't know where this file is being manipulated.
  * EDIT: Wait the pip install will not manipulate the GITHUB repo I'm looking at... disregard?
image

* Running a unit with simple `faraday-telemetry --callsign kb1lqd --nodeid 2` seems to work but it is being saved into a random lib folder....

`faraday-telemetry --callsign kb1lqd --nodeid 2 --database brent.db` produced active telemetry in this folder in C: users .faraday folder. Should we make this more accessible? 

image

So I think this is actually a good PR and with further review looks acceptable, unless I'm missing something @kb1lqc

kb1lqc:
Haha sorry I totally glossed over how that works.

* Yes saving log files to `~/.faraday/lib` is expected and correct, Proxy does the same.
* The config files will be preferentially located in `/etc/faraday/` in the github repo or in `C:/python27/etc/faraday` and lastly in the local folder but 99.99% of the time it will be in the first two

Thanks for checking!

etc/faraday/telemetry.sample.ini
```diff
@@ -4,9 +4,10 @@ PORT=8001
 
 [DATABASE]
 FILENAME=telemetry.db
-SCHEMANAME=db.sql
+SCHEMANAME=telemetrydb.sql
```
kb1lqd, line 7: Good call on proper naming.

kb1lqd (approved):
Looks good, no issues noted during code review and updates installed and appeared to properly run and generate telemetry data/files.

PR #214, opened 2017-06-25
Python Package Support for SimpleUI

kb1lqc:
# Changelog
* setup.cfg for setuptools now includes command line support for `faraday-simpleui`
* setup.cfg also indicates to copy /etc/faraday/simpleui.sample.ini to local directory
* simpleui.py supports reading in loggingConfig.ini and simpleui.ini from local /etc/faraday folder
* Command line arguments added to SimpleUI

## Command Line Arguments
* `--start`
* `--init-config`
* `--callsign`
* `--nodeid`
* `--cmdlocalcallsign`
* `--cmdlocalnodeid`
* `--cmdremotecallsign`
* `--cmdremotenodeid`
* `--flaskhost`
* `--flaskport`
* `--proxyhost`
* `--proxyport`
* `--telemetryhost`
* `--telemetryport`

image

# Python `webbrowser` Module Support

I added support for the [webbrowser](https://docs.python.org/2/library/webbrowser.html) module which automatically reads in the configured Flask `HOST` and `PORT` settings and opens a new window/tab when SimpleUI is run with the `--start` option. SimpleUI must be correctly configured prior to complete functionality and it may be programmed in the same command with `--start`.

simpleuistart


kb1lqd:
* Installed correctly as expected
* Seems to have configured but I'm having a hard time getting proxy data. I'm not sure if it's this commit or something else. I tried prior branches and didn't work properly so I think I might need to restart my computer or such. Will try again later.

kb1lqd (approved):
Installed as expected. Ran proxy, telemetry, and simpleui and all worked. Local and Remote commanding also verified working.

Issue #232, opened 2017-07-04
Add Device Configuration Bitmask Bit - GREEN RX LED


kb1lqd:
## Summary
Due to the PR: https://github.com/FaradayRF/Faraday-Firmware/pull/77

The device configuration package needs an update to allow configuration of this GREEN LED bit within the config boot bitmask.

## Problem Explanation
See Summary

#### Software 
Master software repository.

#### Hardware
REV D1 hardware

# Supporting Information

A GREEN LED version is needed in: https://github.com/FaradayRF/Faraday-Software/blob/master/faraday/deviceconfiguration.py

``` Python
parser.add_argument('--redledtxon', action='store_true', help='Set Faraday radio RED LED during RF transmissions ON')
```

Where the GREEN LED on RX bitmask bit is `BIT2` (00000100)

kb1lqd:
#242 PR closes this issue.

Issue #228, opened 2017-07-04
Add `--start` option control to `faraday-proxy`


kb1lqc:
## Summary
Make command line option `--start` be required to actually start the proxy server. Allow configuration to occur without the `--start` option but exit the program upon completion.

## Problem Explanation
N/A

#### Software 
Faraday Software Master

#### Hardware
N/A

# Supporting Information
This was implemented in device-configuration per PR #219 and it's been very useful. So much so we are rolling it out to all other applications.

kb1lqd:
Implemented in #234

Issue #210, opened 2017-06-22
faraday-proxy has --savelog instead of --save-log


kb1lqc:
## Summary
faraday-proxy command line option of `--savelog` is not consistent with similar options in the prompt. It should be changed to `--save-log` for consistence and commonality to other applications. This is located in the faraday/proxy.py file.

## Problem Explanation
N/A

#### Software 
Software

#### Hardware
N/A

# Supporting Information
N/A

kb1lqc:
It appears this has been addressed and merged.. somehow this ticket doesn't show that c846029 wasn't merged?

![image](https://user-images.githubusercontent.com/331540/32176388-bf5e4dc2-bd44-11e7-97b5-d5dcd4d09c29.png)

kb1lqc:
Yes this ticket was addressed in PR #234 and should be closed.

PR #234, opened 2017-07-05
Add `--start` Option to all Applications (Also Clean Code)

kb1lqc:
Per issues #228, #229, #230, #231, #232, and #233 I've implemented/updated the `--start` option for all applications. This means that `faraday-proxy`, `faraday-telemetry`, `faraday-aprs`, `faraday-simpleui`, `faraday-deviceconfiguration`, and `faraday-simpleconfig` now all require a `--start` option. 

NOTE: This will require update to the quickstart documentation to use the `--start` option before pulling in this PR

Attention to this awesomeness: @reillyeon @el-iso

### DIFF
--- a/faraday/proxy.py
+++ b/faraday/proxy.py
@@ -45,6 +45,11 @@ for location in os.curdir, relpath1, relpath2, setuppath, userpath:
 
 logger = logging.getLogger('Proxy')
 
+# Set werkzeug logging level
+
+werkzeuglog = logging.getLogger('werkzeug')
+werkzeuglog.setLevel(logging.ERROR)
+
 #Create Proxy configuration file path
 proxyConfigPath = os.path.join(path, "proxy.ini")
 logger.debug('Proxy.ini PATH: ' + proxyConfigPath)
@@ -374,6 +379,7 @@ def testdb_read_worker():
 
 # Initialize Flask microframework
 app = Flask(__name__)
+
 
 
 @app.route('/', methods=['GET', 'POST'])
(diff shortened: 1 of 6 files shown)


faraday/proxy.py
```diff
@@ -452,13 +463,13 @@ def proxy():
         # present, create port queue for it and append data to that queue
         try:
             data["data"]
-        except:
+
+        except KeyError:
             logger.error("Error: No 'data' key in dictionary")
             return json.dumps(
                 {"error": "Error: No 'data' key in dictionary"}), 400
         else:
```
reillyeon, line 471: This `else` is unnecessary because the function will return if the `except` block is executed.

faraday/proxy.py
```diff
@@ -452,13 +463,13 @@ def proxy():
         # present, create port queue for it and append data to that queue
         try:
```
reillyeon, line 464: Instead of a try/catch you can use `if "data" in data:` or `if "data" not in data:`.

reillyeon (approved):
This looks good to me once you fix some nits.

kb1lqd:
I created the following BATCH scripts to test this PR:

[faraday-bat.zip](https://github.com/FaradayRF/Faraday-Software/files/1126085/faraday-bat.zip)



It appears that the following programs operated correctly with the `--start` option (non starting program without the `--start` argument but successful program execution when `--start` present).
* proxy
* telemetry
* aprs
* simpleui
* device-configuration
* simpleconfig

kb1lqd (approved):
PR testing passed manual actuation of `--start` on all expected programs. Review of code updates also looks acceptable.

kb1lqc:
This ticket also adsdressed #210

Issue #208, opened 2017-06-19
faraday-proxy Does Not Fail Gracefully if Unconfigured


kb1lqc:
## Summary
After installing faraday-proxy you must configure the `proxy.ini` file with a `faraday-proxy --init-config` and then run `faraday-proxy --callsign xxx --nodeid x --port xxx`. If one just installs faraday and then attempts to run `faraday-proxy` there is an hard failure and the program quits. This should nicely tell the user what is wrong and how to fix it.

## Problem Explanation
Error message:

```
Bryce@BRYCE-SP4 MINGW64 ~/Documents/GitHub/faradayrf-kb1lqc (issue191)
$ faraday-proxy
Traceback (most recent call last):
  File "C:\Users\Bryce\Documents\GitHub\faradayrf-kb1lqc\test\Scripts\faraday-proxy-script.py", line 11, in <module>
    load_entry_point('faraday', 'console_scripts', 'faraday-proxy')()
  File "c:\users\bryce\documents\github\faradayrf-kb1lqc\test\lib\site-packages\pkg_resources\__init__.py", line 561, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "c:\users\bryce\documents\github\faradayrf-kb1lqc\test\lib\site-packages\pkg_resources\__init__.py", line 2649, in load_entry_point
    return ep.load()
  File "c:\users\bryce\documents\github\faradayrf-kb1lqc\test\lib\site-packages\pkg_resources\__init__.py", line 2303, in load
    return self.resolve()
  File "c:\users\bryce\documents\github\faradayrf-kb1lqc\test\lib\site-packages\pkg_resources\__init__.py", line 2309, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "c:\users\bryce\documents\github\faradayrf-kb1lqc\faraday\proxy.py", line 216, in <module>
    configureProxy(args, proxyConfigPath)
  File "c:\users\bryce\documents\github\faradayrf-kb1lqc\faraday\proxy.py", line 173, in configureProxy
    config.set(unit, 'BAUDRATE', args.baudrate)
  File "C:\python27\Lib\ConfigParser.py", line 396, in set
    raise NoSectionError(section)
ConfigParser.NoSectionError: No section: 'UNIT0'
```

#### Software 
Faraday software

#### Hardware
All

# Supporting Information
N/A

kb1lqd:
#235 closes this.

PR #235, opened 2017-07-05
Fix Proxy Configuration Exceptions

kb1lqc:
Per #208 this PR now catches exceptions when:
- `--init-config` has not been run and no proxy.ini exists

Since this PR is based on PR #195 (so please pull that in first or else this includes all of #195 as well...) it also catches UARTstack exceptions when the UART port is mis-configured. Overall this is a quick fix to a pesky issue that makes Proxy a bit cleaner to use.

@el-iso  @reillyeon @kb1lqd

### DIFF
--- a/faraday/proxy.py
+++ b/faraday/proxy.py
@@ -213,7 +213,16 @@ def configureProxy(args, proxyConfigPath):
 # Initialize and configure proxy
 if args.init:
     initializeProxyConfig()
-configureProxy(args, proxyConfigPath)
+
+# Attempt to configure proxy
+try:
+    configureProxy(args, proxyConfigPath)
+
+except ConfigParser.NoSectionError as e:
+    # Possible that no configuration file found
+    logger.error('Proxy configuration file error!')
+    logger.error(e)
+    sys.exit(0)
 
 # Load Proxy Configuration from proxy.ini file
 proxyConfig = ConfigParser.RawConfigParser()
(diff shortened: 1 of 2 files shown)


faraday/proxy.py
```diff
@@ -219,7 +219,17 @@ def configureProxy(args, proxyConfigPath):
...
-configureProxy(args, proxyConfigPath)
+
+# Attempt to configure proxy
+try:
+    configureProxy(args, proxyConfigPath)
+
+except ConfigParser.NoSectionError as e:
+    # Possible that no configuration file found
+    logger.error('Proxy configuration file error!')
+    logger.error('Did you remember to --init-config?')
```
kb1lqd, line 230: @kb1lqc I fixed and pushed a typo here. You spelled "remember" incorrectly.

faraday/uart/layer_4_service.py
```diff
@@ -204,7 +204,7 @@ def uart_layer_receive_link(self):
             pass
 
     def Abort(self):
-        logger.error("Abording layer 4 faraday_uart_object")
+        logger.error("Aborting layer 4 faraday_uart_object")
```
kb1lqd, line 207: Nice!

kb1lqd (approved):
@kb1lqc I fixed a small typo and pushed it. All in all it looks good and seems to operate as expected in my testing. We should make another PR to bring this into the other programs too.

PR #238, opened 2017-07-07
GREEN RX config bitmask bit added for issue#232

markwiem:
Added logic to solve issue #232 -- disclaimer, I do not own a FaradayRF (yet) so this code has unfortunately not been tested. I've vetted the changes to the best of my ability.   

Please feel free to provide feedback of any errors made or suggestions for improvements.


kb1lqd:
Looks like I'm not the only one burning the midnight oil. Thanks @markwiem ! I'll check this out tomorrow and provide any feedback if needed or go ahead and merge.

kb1lqd:
@markwiem If you check our PEP8 travis-ci test that failed there are two line errors. Can you fix them prior to my review?

https://travis-ci.org/FaradayRF/Faraday-Software/builds/251073179#L505


p.s. Our big move this week to command line tools and such will allow us to bring in larger automated system testing in the future.

markwiem:
@kb1lqd apologies! My editor acted up right before the PR, and of course I missed that some syntax fell off the line. 

Looking forward to your review.

kb1lqc:
@kb1lqd does this require newer firmware than the one we installed performing Data testing last night?

kb1lqc (changes requested):
First observation is this requires additions to faraday_config.sample.ini in `etc/faraday` to include the `GREENLEDRX` option due to the error below:

image

Check out the REDLEDTX option:
https://github.com/FaradayRF/Faraday-Software/blob/master/etc/faraday/faraday_config.sample.ini#L2

kb1lqc:
I can confirm that with a manually added `GREENLEDRX` option to the `faraday_config.ini` file This logic correctly changes the bit value with `--greenledrxon` and `--greenledrxoff` 👍

kb1lqc:
Oops not obvious but also noticed that a line similar to this (with correct bit) needs to be added to `simpleconfig.py` in /faraday:

`logger.info("GREENLEDRX: {0}".format(bootbitmask[?]))`

kb1lqc:
OK tested this on my hardware and it didn't work. However I think I may be running an older firmware than is supported here. Need to head into work. @kb1lqd maybe give it a shot today with known good firmware for this PR of software and @markwiem thanks for the contribution. The two comments I have on items that need updating were NOT obvious without being told/having hardware. Thanks for the help so far!

kb1lqd:
OK Starting review, I'm not sure why this PR is showing the prior commits in @kb1lqc and @kb1lqd PR's from last night as new changed files...

kb1lqd:
It seems that this `eightBitListToInt()` function has reversed the bits and is confusing. The correct allocations are 

[BIT7, BIT6, BIT5, BIT4, BIT3, BIT2, BIT1, BIT0]

BIT0 = Flash Programs
BIT1 = REDLEDTXENABLED
BIT2 = GREENLEDRXENABLED

But this list has the order swapped:

image

This causes the configbootbitmask to be 35:

image

Which in binary clearly shows the BIT2 for GREEN LED in the BIT5 location :(

image


Green LED BIT is correct but the other two are at list index 6 and 7 which are BIT1 and BIT0 respectively. This is non-intuitive but I see that the function that converts this to an INT just joins the list and thus index 7 of the list is BIT0. The `eightBitListToInt()` docstring should clearly explain this expectation and/or swap the ordering to match expected.

**EDIT:**
Ahh, I see in https://github.com/FaradayRF/Faraday-Software/pull/238/files#diff-daca2c9d995209f5746fc52745a9463bR143 that this is handled by reversing the order _just for display when reading data..._ This is confusing, we should always keep the bits in the correct order as long as possible and not need to do this.

faraday/deviceconfiguration.py
```diff
@@ -68,6 +68,8 @@
 parser.add_argument('--nodeid', type=int, help='Set Faraday radio nodeid')
 parser.add_argument('--redledtxon', action='store_true', help='Set Faraday radio RED LED during RF transmissions ON')
 parser.add_argument('--redledtxoff', action='store_true', help='Set Faraday radio RED LED during RF transmissions OFF')
+parser.add_argument('--greenledrxon', action='store_true', help='Set Faraday radio GREEN LED during RF reception ON')
```
kb1lqd, line 71: Looks good.

faraday/deviceconfiguration.py
```diff
@@ -222,6 +224,10 @@ def configureDeviceConfiguration(args, deviceConfigurationConfigPath, faradayCon
         fconfig.set('BASIC', 'REDLEDTX', 1)
     if args.redledtxoff:
         fconfig.set('BASIC', 'REDLEDTX', 0)
+    if args.greenledrxon:
```
kb1lqd, line 227: Looks good.

@kb1lqc I just realized that all these functions don't seem to have a default or error if BOTH options are in the arguments...?

faraday/simpleconfig.py
```diff
@@ -143,6 +143,7 @@
 bootbitmask.reverse()
 logger.info("UNITCONFIGURED: {0}".format(bootbitmask[0]))
 logger.info("REDLEDTX: {0}".format(bootbitmask[1]))
+logger.info("GREENLEDRX: {0}".format(bootbitmask[5]))
```
kb1lqd: This is incorrect and should be `bootbitmask[2]`

faraday/deviceconfiguration.py
```diff
@@ -230,7 +236,9 @@ def configureDeviceConfiguration(args, deviceConfigurationConfigPath, faradayCon
     # Create configuration boot bitmask integer
     bootmask = [0] * 8
     redledtx = fconfig.get('BASIC', 'REDLEDTX')
+    greenledrx = fconfig.get('BASIC', 'GREENLEDRX')
     unitconfigured = fconfig.get('BASIC', 'UNITCONFIGURED')
+    bootmask[2] = greenledrx
```
kb1lqd: This is incorrect and should be `bootmask[5]`.

The root cause of this error is confusing eightBitListToInt() ordering functionality but this should be clarified/updated in a future PR.

kb1lqd (changes requested):
@markwiem 

Please make the following changes:

* https://github.com/FaradayRF/Faraday-Software/pull/238/files#diff-daca2c9d995209f5746fc52745a9463bR146
  * `bootbitmask[5]` should be `bootbitmask[2]`
* https://github.com/FaradayRF/Faraday-Software/pull/238/files#diff-9fedb24251951e38ac8e7610040dcd75R241
  * `bootmask[2]`. should be `bootmask[5]`.

With these updates I have confirmed that the GREEN LED does blink when RF data is received!

Also,

@kb1lqc Do you know why this is saying all the `data.py` merge files from prior PR are "new"?

markwiem:
@kb1lqd , data.py merge files from prior PR being marked "new" is likely my fault. I was following the workflow at https://gist.github.com/Chaser324/ce0505fbed06b947d962#keeping-your-fork-up-to-date at performed `git merge upstream/master` (in case there had been any early morning commits). I'm new to the forking process and keeping track of upstream changes. Please let me know if I erred in any way and feel free to dismiss this PR if it's a liability to `master`.

faraday/deviceconfiguration.py
markwiem: Correct, I was confused by `bootbitmask.reverse()`. Thanks for clarifying, edit made.

faraday/simpleconfig.py
markwiem: Correct, I was confused by `bootbitmask.reverse()`. Thanks for clarifying, edit made.

kb1lqd:
Can confirm @markwiem that https://github.com/FaradayRF/Faraday-Software/pull/238/commits/3a85ecc84c17b6292071af0d29806f4242af2eef has fixed the bitmask errors and the GREEN LED RX LED is toggled ENABLED/DISABLED with the configuration program.

I'd like to see if we can get the other files `data.py` and such to clear. I understand learning the forking method, I'm still learning it myself.

@markwiem and you do https://help.github.com/articles/syncing-a-fork/ and see if you successfully pull in the merges?

Might even try a `git rebase` as well. Seems OK when I tried this with faraday master repo although might need to --skip some of the patches as I did. I'm still new to this as well.

PR #247, opened 2017-07-16
Update Start Tutorials With --start Option

kb1lqc:
We realized that the Tutorials were missing several spots where knowing that you need to use `--start` to start up a server was crucial. These are now fixed and it should be clear how to start the desired program server when relevant.

This should be a relatively simple update to tutorials. Please read and let me know if I missed anything!


kb1lqd (approved):
Approved but please add the following:

Add a comment near a relevant portion of the tutorials stating that when pip installed in editable mode you may need to run commands from within the main directory.

kb1lqc:
@kb1lqd we actually already have that warning in the "Installing in Pip Editable Mode" section on `installing-sofware.md`

image

Merged, thanks @kb1lqd  and @jbemenheiser!

Issue #244, opened 2017-07-11
Faradaybasicproxyio has hardcoded IP/hostnames


kb1lqc:
## Summary
The faraday/proxyio/faradaybasicproxyio.py file with serves up our GET and POST functions for obtaining data RESTfully has hardcoded locahost values of `requests.post("http://127.0.0.1:" + str(self.FLASK_PORT) +.........` for example. This is not good and we should allow the user to send a specific IP/Hostname and default to localhost if not provided.

## Problem Explanation
N/A

#### Software 
Faraday-Software master

#### Hardware
N/A

# Supporting Information
This is stopping us from remotely connecting to radios.

kb1lqc:
Fixed in #256 so closing!

PR #256, opened 2017-08-13
Issue244: Adding in Dynamic Proxy, Telemetry, and SimpleUI URLs

kb1lqc:
Per #244 this provides Proxy, Telemetry, and SimpleUI with the ability to use url's defined in the respective configuration files for Flask urls used to obtain telemetry and command units.

Todo:

- [x] Check that all configuration sample files have appropriate defaults
- [x] Catch expections when all new configuration elements fail to import (i.e. "REPLACEME" imported as an int.
- [ ] Test Proxy over internet
- [ ] Test Telemetry over internet
- [x] Consider configuration to disabled proxy POST operation for a first-attempt at somewhat safe hosting 

online and not allows stations to transmit data from anyone... licensed or not.

## Changelog
- Updated faradaybasicproxyio GET function with dynamic host address parameter
- Updated faradaybasicproxyio GET function with .format() string manipulation
- Updated faradaybasicproxyio POST function with dynamic host address parameter
- Updated faradaybasicproxyio POST function with .format() string manipulation
- Updated Proxy, Telemetry, and Simpleui to use `proxyhost` value or standard `host` in the flask section to set host of proxy server
- Updated javascript and HTML templates to import host/port information about proxy from Python
- Updated SimpleUI to check for command stations and catch exceptions when misconfigured.
- Added a very simplistict proxy flag to disabled adding anything to the POST queue.
- Updated proxy, telemetry, and simpleui to include relevant `disabletx`, `proxyhost` configuration values from command line


kb1lqc:
OK I tried to check sending Faraday proxy over an open internet connection to the world using a VPN. This did not apparently work. Maybe I did something wrong? @hdkmike have any ideas?

I essentially wanted to have flask server to "0.0.0.0" port 8000 which should set it to accept any connection over the internet (with debug=false set in Flask). I could connect over local router IP fine but the second I tried binding to a public IP nothing worked. I also checked my router port forwarding which should have been transferring all port 8000 activity to my desktop computer IP.

Overall I think I'm going to move on as this PR isn't specifically about serving over the internet, it's about cleaning up the IP/Port hard-coding. I was hoping this would be a two-for-one deal. Apparently not.

kb1lqc:
@kb1lqd want me to pull this in without your review?

faraday/proxy.py
```diff
@@ -59,6 +59,7 @@
 parser.add_argument('--test-nodeid', dest='testnodeid', type=int, help='Set Faraday test mode nodeid')
 parser.add_argument('--test-rate', dest='testrate', default=1, type=int, help='Set Faraday test mode rate')
 parser.add_argument('--payload-size', dest='payloadsize', default=39, type=int, help='Set Faraday data mode payload size')
+parser.add_argument('--disabletx', type=int, default=0, help='Disable proxy from POSTing to Faraday thus preventing transmissions')
```
kb1lqd: Not sure why you want this but not stopping.

kb1lqd (approved):
I ran the PR and proxy, telemetry, APRS, and simpleui worked as expected. Local commanding with simplyUI also worked, I didn't have the correct unit ID set but sending remote command caused the local unit to TX. I did not intentionally run alternate host addresses but saw that the default localhost worked. A quick review of the code looks like the updates are OK. 

I don't know why you're putting in the --disabletx option but sounds like you needed it and see a reason. I can see so that you ensure no transmissions but looks like you also block any local POST'ing that would NOT result in a transmission (i.e. local LED command).

kb1lqc:
@kb1lqd yeah that was rushed, I removed it in 0c5760f. Tested and all seems to be working. Next up on the whole project management tasks is unit testing. One day we can have nice things. One day :)

Issue #266, opened 2017-09-04
Implement APRS BASE91 Comment Telemetry


kb1lqc:
## Summary
8 Bit telemetry is, well, antiquated. This issue ticket will implement BASE91 comment telemetry into the `faraday-aprs` package. The result will be a full 12 bit value for each telemetry field.

## Problem Explanation
Some developers added on the ability for APRS-IS and websites such as APRS.fi to support BASE91 encoded telemetry in comments. This allows up to 16 bit accuracy for each telemetry field.

#### Software 
>Please include the branch or commit if known regarding what software was used if applicable

#### Hardware
>Let us know about what hardware you are using if applicable. This includes:
>- Faraday radio serial numbers
>- Computer operating system

# Supporting Information

- https://aprs.fi/doc/guide/aprsfi-telemetry.html
- http://he.fi/doc/aprs-base91-comment-telemetry.txt

kb1lqc:
Installing `base91` from pip: https://pypi.python.org/pypi/base91/1.0.1

kb1lqc:
Oh sweet Jesus... this stumped me for a few minutes

```
3.4 The telemetry sequence counter and telemetry channels are encoded using
the Base91 encoding, as it is presently used in Compressed APRS position
packets, the altitude extension, and the DAO extension. Two bytes are
transmitted for the sequence counter and each of the channels, giving over
13 bits of resolution (values 0 to 8280). Please note that APRS uses a
different definition of Base91 than the internet standard Base91 - see
APRS101.PDF for details.
```

> Please note that APRS uses a different definition of Base91 than the internet standard Base91 - see APRS101.PDF for details.

🤔

kb1lqc:
Well even more reason to move to `aprslib`, implements base91 APRS flavor https://github.com/rossengeorgiev/aprs-python/blob/master/aprslib/base91.py

![standards](https://user-images.githubusercontent.com/331540/30040705-c2a786ca-9197-11e7-8e37-8b7551315c4c.png)

kb1lqc:
Alright I got it. It turns out that `aprslib` (or [aprs-python](https://github.com/rossengeorgiev/aprs-python)) has a BASE91 encoder/decoder. 

I completely missed that you could specify width with `aprslib` BASE91... woops. I thought there was an error

kb1lqc:
Ran code overnight to get 12 bit ADC data into APRS-IS!

[...]

PR #269, opened 2017-09-08
Aprs telem

kb1lqc:
# Changelog
- Imports aprslib and uses base91 functionality. Please note that this is note a standard BASE91 implementation and is a custom APRS version of base91. Qeue [XKCD](https://xkcd.com/927/).
- Removes use of `sendTelemetry` function as it is no longer needed
- Only sends scaling values on first transmission to reduce unnecessary packet transmissions. This should be revisited later but seems good for now.
- Sends a `telemetrySequence` number that is unbounded to `sendPosition` as BASE91 comment telemetry can support up to 16 bits of counting and this should be fine for the forceable future :)
- `sendPosition` implements [BASE91 Comment Telemetry](http://he.fi/doc/aprs-base91-comment-telemetry.txt) with ADC0, ADC1, ADC3, ADC6, BOARDTEMP, and GPIO bits per #266.

Raw packets which end in BASE91 comment telemetry 
image

12 bit ADC data looks very nice!
image

### DIFF
--- a/faraday/aprs.py
+++ b/faraday/aprs.py
@@ -17,6 +17,7 @@ import os
 from time import sleep
 import sys
 import argparse
+from aprslib import base91
 
 from classes import helper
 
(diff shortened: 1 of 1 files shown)


kb1lqc:
@el-iso you have time to give this a review?! Just let me know if not. Totally fine but could use the extra eyes on it!

Issue #262, opened 2017-09-01
Create an "it just works" Faraday Raspberry Pi base station project implementation.


kb1lqd:
Create and document a linux/Rapberry Pi integration of telemtry, GUI, command, and APRS functionality to allow easy and intuitive Faraday remote telemetry and command projects.

Features:
* 24/7 telemetry logging
* 24/7 APRS server uploading
* Easy start scripts
* Improved unit commanding interface (through Open MCT #259 ?)
* Setup instructions to easily SSH/RDP into Rpi
  * Rpi autoconnects to wifi hotspot
  * Rpi IS the wifi hotspot
* Example hardware setup

Issue #201, opened 2017-05-25
Faraday Range Testing


el-iso:
## Summary
I am planning to do some range testing of Faraday with my friend Kevin (KA9UBD). @kb1lqd Suggested that I post an issue to discuss the best methods to setup the test and how to collect useful information for the overall Faraday project.

## Problem Explanation
We are planning to setup one faraday node at a high point in our area with a base station antenna. The other node will be in our car plugged into an internet connected computer running proxy, telemetry, and APRS. Upon setting up the base station node we will drive until we lose communications with the base station recording telemetry the whole way. 

## Questions
* Brent ( @kb1lqd ) suggested that testing transmissions at varying power levels would be beneficial. What would be the best way to control that in the code? It seems like this might require putting a computer at the base stations as well?
* What other modification to code or our testing setup would be beneficial?

This will be our first real project with Faraday other than when we did setup, so any suggestions or advice on how best to run the test would be very appreciated.

kb1lqc:
@el-iso thanks for making this test public! Here are some thoughts:

- Power levels are currently only set by firmware update so it's not the easiest to change on the fly. Changing power levels could be interesting but also take one step at a time and possibly just focus on highest power for now.
- Make sure to test everything a day or so prior, make sure you're ready to go from a Faraday and APRS perspective. Morning of issues suck.
- Keep the [radio horizon in mind](https://www.everythingrf.com/rf-calculators/line-of-sight-calculator), it will absolutely reduce your range.
- If you can use a better antenna than the whip on the car unit then that's awesome but not required. The Faraday whip antennas are great for whip antennas but still don't compare well against a great base station antenna or yagi.

Another thought though definitely on a whim is that you could have something similar to [hello world](https://github.com/FaradayRF/Faraday-Software/blob/master/Tutorials/start/hello-world.py) but modified for RF to command from the base station to remote unit to blink the LEDs. Then as you're driving around you can see visually when the radio stops receiving the commands from base station.

Hope this helps!

el-iso:
Thanks! @kb1lqc 

We were planning to try for a test today, but it looks like we might be getting some bad weather so were going to postpone. Kevin Price and I are going to work on some prep and testing today though, so I will keep all these in mind! Especially the "hello world" type program that would be useful. 

As far as results of the any range testing we do I'll plan to write a report of what we see. Would you like me to share the contents of the database after the fact?

Thanks again!

kb1lqd:
[...]

Issue #199, opened 2017-05-07
Long Duration APRS Failed - [Errno 32] Broken pipe


kb1lqd:
## Summary
During a long duration APRS telemetry run of only a local faraday device using a Raspberry Pi 3 the APRS server stopped due to `[Errno 32] Broken pipe`.

## Problem Explanation

I allowed a single Faraday (local) to log telemetry and upload to APRS but the APRS stopped after a few days. Investigating into why is stopped I noticed that the tmux screen running `aprs.py` was showing errors:

```BASH
2017-05-07 05:53:43,321 - APRS - INFO - Tracking 1 Faraday stations...
2017-05-07 05:53:43,322 - APRS - ERROR - SendPosition
2017-05-07 05:53:43,323 - APRS - ERROR - [Errno 32] Broken pipe
2017-05-07 05:53:43,324 - APRS - ERROR - Sendtelemetry
2017-05-07 05:53:43,324 - APRS - ERROR - [Errno 32] Broken pipe
2017-05-07 05:53:43,325 - APRS - ERROR - SendTelemLabels
2017-05-07 05:53:43,326 - APRS - ERROR - [Errno 32] Broken pipe
2017-05-07 05:53:43,327 - APRS - ERROR - SendParameters
2017-05-07 05:53:43,328 - APRS - ERROR - [Errno 32] Broken pipe
2017-05-07 05:53:43,329 - APRS - ERROR - SendEquations
2017-05-07 05:53:43,329 - APRS - ERROR - [Errno 32] Broken pipe
```

Simply CTRL+C stopping `aprs.py` and restarting the `aprs.py` program fixed the issue and aprs.fi shows the Farday unit.

#### Software 

```BASH
commit 2ba04caff9affb0376eb4997a92933ba993c56a5
Merge: 71086e8 6340882
Author: Bryce Salmi <bryce@faradayrf.com>
Date:   Sun Apr 30 13:02:29 2017 -0700

    Merge pull request #196 from FaradayRF/kb1lqc-patch-1

    Quick Fix To Update Proxy.py To sys.exit(1)
```

#### Hardware
* REV D1 Faraday units
* Raspberry Pi 3

# Supporting Information

Below is an image of the noted error from the APRS TMUX screen:

![image](https://cloud.githubusercontent.com/assets/339144/25778308/e6f00d06-32ae-11e7-9b49-412d2ce6d71e.png)

Restarting only APRS made APRS properly update to the APRS-IS system:

[...]

Issue #190, opened 2017-04-28
Proxy/Telemetry/APRS Error Long Duration Raspberry Pi 3


kb1lqd:
## Summary
While letting a Raspberry Pi 3 run the noted commit run for ~1.5 days it seems that APRS.py errored. Telemetry and Proxy remained running although after a restart of APRS.py only ARPS uploading restarted for only 1 of two units. 1 Unit is local while the other is remote (RF).

## Problem Explanation

This long duration test of proxy, telemetry, and APRS scripts ran for over a day but failed due to what looks like unexpected data/format from unit `KB1LQD-25`. The error crash APRS.py, I was logging proxy logs and telemetry logs. Restarting the APRS.py script refreshed `KB1LQD-2` but not `KB1LQD-15`. 

#### Software 

Running Master:

commit 3d91490717f4c8782f8ebe86c07c8b873fa271e3
Merge: 421aa2f 84c67be
Author: Reilly Grant <reillyeon@qotw.net>
Date:   Mon Apr 24 00:23:35 2017 -0700

    Merge pull request #174 from FaradayRF/kb1lqc-proxylink-2

    Fixed readme.md Proxy folder links
#### Hardware
* Faraday REV D1 - SN:18 (Local)
* Faraday REV D1 - SN:13 (Remote - powered by external USB HUB)

# Supporting Information

## Logs

* Proxy = `logs.db`
* Telemetry = `telemetry.db`

[Logs.zip](https://github.com/FaradayRF/Faraday-Software/files/963578/Logs.zip)

### Proxy
![image](https://cloud.githubusercontent.com/assets/339144/25516699/e0f32858-2ba0-11e7-821b-b9240d60b6e3.png)

### Telemetry
![image](https://cloud.githubusercontent.com/assets/339144/25516701/e44ac3c6-2ba0-11e7-993d-1178f228f124.png)

### APRS

![image](https://cloud.githubusercontent.com/assets/339144/25516689/d5df1d5a-2ba0-11e7-851a-a3201d4774dd.png)

APRS Stopped working randomly after many hours left running (~1.5 days). It seems like only APRS died.

I restarted APRS:
![image](https://cloud.githubusercontent.com/assets/339144/25516643/92ffc1ce-2ba0-11e7-909f-9ada33361c02.png)

A little longer later…

![image](https://cloud.githubusercontent.com/assets/339144/25516659/a8a30130-2ba0-11e7-8e48-92bc7ae81a2e.png)

[...]

PR #271, opened 2017-09-10
Make Faraday-APRS Persistent Across Network Failures

kb1lqc:
# BASE91 APRS Comment Telemetry
[aprs.fi](www.aprs.fi) and a few other services support BASE91 telemetry transmission. This allows up to 13 bits of ADC telemetry per channel. This PR pulled in what was #269 so here is the  information relevant to that PR. @reillyeon beat @el-iso to the punch on there. I appreciate the help from both!

## Changelog
- Imports aprslib and uses base91 functionality. Please note that this is note a standard BASE91 implementation and is a custom APRS version of base91. Qeue [XKCD](https://xkcd.com/927/).
- Removes use of `sendTelemetry` function as it is no longer needed
- Only sends scaling values on first transmission to reduce unnecessary packet transmissions. This should be revisited later but seems good for now.
- Sends a `telemetrySequence` number that is unbounded to `sendPosition` as BASE91 comment telemetry can support up to 16 bits of counting and this should be fine for the forceable future :)
- `sendPosition` implements [BASE91 Comment Telemetry](http://he.fi/doc/aprs-base91-comment-telemetry.txt) with ADC0, ADC1, ADC3, ADC6, BOARDTEMP, and GPIO bits per #266.

Raw packets which end in BASE91 comment telemetry 
image

12 bit ADC data looks very nice!
image

# Persistent APRS-IS Connection Across Network

This PR is based on #269 as it was branched from that PR repository and aims to improve reliability when connecting to APRS-IS servers across a network. As we've seen, often a few minutes to a few hours after starting APRS logging a network event will cause  socket ERRNO 10054 or similar and `faraday-aprs` could not recover. This was annoying at best and at worst rendered the program not terribly useful.

Mind helping me out @reillyeon, if no time I can see if @kb1lqd has some free time in the near future :)

## Changelog
- Created a `sendAPRSPacket` function to consolidate sending of data to the socket and exception catching
- Reconnects after a IOERROR in 2 second loop attempts

## Issue Tickets
This PR addresses issues related to the following tickets Faraday-Software #119, Faraday-Software #190, Faraday-Software #199, Faraday-Software #201, Faraday-Software #262. A few of these tickets can likely be completely closed as a result of this testing and eventual PR!


faraday/aprs.py
```diff
@@ -103,6 +104,11 @@ def aprs_worker(config, sock):
     # Local variable initialization
     telemSequence = 0
 
+    #dev
+    sock = ''
```
reillyeon: This line is unnecessary.

faraday/aprs.py
```diff
@@ -114,11 +120,21 @@ def aprs_worker(config, sock):
...
-        sendParameters(stationData, sock)
-        sendEquations(stationData, sock)
+        if not conn:
+            sock = connectAPRSIS()
+        try:
+            conn = sendPositions(telemSequence, stationData, sock)
+
+            # Just send labels, Parameters, and Equations every 10th loop
+            if telemSequence % 10 == 0:
+
```
reillyeon: nit: Extra blank line.

faraday/aprs.py
reillyeon: nit: This statement is unnecessary.

faraday/aprs.py
```diff
@@ -317,11 +382,8 @@ def sendPositions(stations, socket):
 
                 logger.debug(positionString)
 
-                try:
-                    socket.sendall(positionString)
-
-                except IOError as e:
-                    logger.error(e)
+                status = sendAPRSPacket(socket, positionString)
+                return status
```
reillyeon: `return sendAPRSPacket(socket, positionString)`

Here and below.

faraday/aprs.py
```diff
@@ -693,14 +714,17 @@ def connectAPRSIS():
...
             except IOError as e:
+                # Close socket and setup for next connection attempt
                 logger.error(e)
+                aprssock.close()
+                aprssock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
 
             else:
                 logger.info("Connection successful!")
                 return aprssock
                 break
```
reillyeon: Not this change, but `break` after `return` does nothing.

faraday/aprs.py
```diff
@@ -762,7 +786,8 @@ def main():
     """
 
     logger.info('Starting Faraday APRS-IS application')
-    sock = connectAPRSIS()
+    #sock = connectAPRSIS()
+    sock = ''
```
reillyeon: Why pass `sock` to `aprs_worker` at all if it calls `connectAPRSIS()` itself?

reillyeon (changes requested):

faraday/aprs.py
kb1lqc: Removed, `conn=False` now used in the initialization just above to force it to initialize on first loop. Thanks for catching this!

kb1lqc: ?

faraday/aprs.py
kb1lqc: Removed

faraday/aprs.py
kb1lqc: All `return status` lines replaced with obviously better `return sendAPRSPacket()`. Thanks for observing this unnecessary addition of variables!

faraday/aprs.py
kb1lqc: Removed

faraday/aprs.py
kb1lqc: Good point. I used to need this but with the current implementation of the socket connections I do not. Removed!

faraday/aprs.py
```diff
@@ -693,14 +699,16 @@ def connectAPRSIS():
...
+                aprssock.close()
+                aprssock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
 
             else:
                 logger.info("Connection successful!")
                 return aprssock
-                break
 
-            sleep(10)  # Try to reconnect every 10 seconds
+            sleep(2)  # Try to reconnect every 10 seconds
```
reillyeon: Update the comment?

faraday/aprs.py
```diff
@@ -268,11 +302,37 @@ def sendPositions(stations, socket):
...
+        b91c = base91.from_decimal(station["ADC3"], 2)
+        b91d = base91.from_decimal(station["ADC6"], 2)
+        b91e = base91.from_decimal(station["BOARDTEMP"], 2)
+        b91f = base91.from_decimal(ioList, 2)
+
+        b91Tlm = "|{0}{1}{2}{3}{4}{5}{6}|".format(b91seq, b91a, b91b, b91c, b91d, b91e, b91f)
+
+        # add telemetry to comments
+        comment = comment + b91Tlm
+        altComment = altComment + b91Tlm
```
reillyeon: comment += b91Tlm
    altComment += b91Tlm

`a += b` is more efficient than `a = a + b`.

reillyeon (approved):

faraday/aprs.py
kb1lqc: Updated

faraday/aprs.py
kb1lqc: Updated

kb1lqc:
@reillyeon if you're going to help pull in the code that is effectively PR #269 then I should bring over the changelong here too. I appreciate the help on these if you are doing both!

reillyeon:
PR #269 plus the changes here look good to me.

kb1lqc:
Learn something new everyday... By pulling in #271 github automatically merged #269 since it was based on the same branch. This is awesome and in retrospect obvious! This means that as we progress through a big task we can commit mini PR's on it and still have one big PR at the end. If one PR that is a combination of a few smaller PR's in merged then we no longer have to worry about those. I like that!

@kb1lqd for awareness.

kb1lqd:
Whoa thanks all and now I learned something new too, very cool PR trait!

Issue #276, opened 2017-10-21
Device Configuration Erroring During Runtime


kb1lqd:
## Summary
As of the last few months the device configuration software used to updated flash memory configuration of Faraday errors and does not update the configuration.

## Problem Explanation
It seems that something in the device configuration program errors, possibly due to INI file issues due to the error logging messages.

#### Software 
Current Master as of ticket date.

#### Hardware
Faraday REV D1.

# Supporting Information

`faraday-devieconfiguration --init-config`
![image](https://user-images.githubusercontent.com/339144/31853364-f9b13f94-b63b-11e7-8eb8-51c00c78eae7.png)

`faraday-deviceconfiguration --init-faraday-config`

![image](https://user-images.githubusercontent.com/339144/31853410-a1fde5d0-b63c-11e7-9218-885b0f504bfe.png)

OK I uninstalled and reinstalled in editable mode the init's are working:

![image](https://user-images.githubusercontent.com/339144/31853543-0f97e6a2-b63f-11e7-8fb7-6b48b74a0c9a.png)

kb1lqd:
@kb1lqc I totally forgot where these files above are being copied from. I think the sample files are missing and that's why device configuration is failing to initialize and work. Can you confirm?

EDIT: NVM Found them in etc which I recall is correct. I figured out that something was wrong with the installation ( I recall doing it...) but after uninstall and reinstalling in editable mode the init's work.

kb1lqd:
OK So I got proxy and device configuration running (realized I needed to run commands in the root folder only in editable mode too) and I ran into this error when trying to run simple config:

![image](https://user-images.githubusercontent.com/339144/31853605-1323d26c-b640-11e7-81f4-7f372d6625a3.png)

kb1lqd:
Ohhh wow I think I found something @kb1lqc 

When I init a new device configuration I see the default values:

![image](https://user-images.githubusercontent.com/339144/31853655-069e8c8e-b641-11e7-875e-6fb55e0dc03c.png)


When I update basic information I see the callsign updated with a "?" prepended...

`$ faraday-deviceconfiguration --proxycallsign "kb1lqd" --proxynodeid 1 --callsign "kb1lqd" --nodeid 1 --gpsbootoff`

![image](https://user-images.githubusercontent.com/339144/31853667-3c3e58b0-b641-11e7-9352-58db4c0b66e5.png)

kb1lqd:
[...]

PR #277, opened 2017-10-22
Issue276

kb1lqc:
Fixes the POST/GET call errors observed in #276 which prevented configuration of Faraday Radios.


kb1lqd (commented):
Tried on my computer/unit and all worked fine! Simple script for programming works. Nice job.

Issue #218, opened 2017-06-30
Python Requirements.txt Should Default to Newest Versions Where Possible


kb1lqc:
## Summary
I noticed that we are forcing Python packages such as `Requests` version 2.12.5 when version 2.18.1 is available. While there could be some issues with newer packages I think we can safely default to the latest versions after the currently indicated versions (what Faraday-Software was initially built with).

## Problem Explanation
I don't like seeing us forcing older code than necessary. Especially as we start automating our testing we should be able to automatically detect when a newer version of a package Faraday-Software uses breaks the build. We should update `requirements.txt` to default to "=>" for most python modules.

@kb1lqd @el-iso @reillyeon 

#### Software 
All platforms running our python code.

#### Hardware
N/A

# Supporting Information

- https://pip.readthedocs.io/en/1.1/requirements.html
- https://github.com/FaradayRF/Faraday-Software/blob/master/requirements.txt

kb1lqc:
Addressed in #281

PR #281, opened 2017-10-31
Updated requirements with >= per #218

kb1lqc:
# Changes

Literally just made requirements document now call out >= for packages. This assumes all packages are backwards compatible which seems like a reasonable assumption for now. Closes out #218


kb1lqc:
This actually passes pytest:
image

Travis CI seems to have just failed. I don't feel need to restart PR.

PR #282, opened 2017-10-31
Added E722 to flake8 ignore list per travis CI

kb1lqc:
Travis CI showed E722 failing which is weird because nothing changed about it.. This also passes without the ignore for E722 when run locally with pytest. We should however ignore it because we haven't fully checked for this and one day will. I don't really understand how this was missed prior haha.

E722 is the use of a bare exception. I avoid it in most places though there are a few spot in my code I still use it. Some of our very old code is full of them however.

@reillyeon @kb1lqd

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [tool:pytest]
 addopts = --flake8
-flake8-ignore = E262 E265 E266 E402 E501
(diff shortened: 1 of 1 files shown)


reillyeon:
With your recent change to `requirements.txt` it is now possible that Travis CI will be running a newer version of the linting tools than you are locally. This can lead to inconsistencies like this.

